PR #9, opened 2022-12-30
Delete fill_types

Akuli:
fill_types was a weird compilation step, and is unnecessary now that we have a CFG.


src/build_cfg.c
```diff
@@ -36,61 +50,213 @@ static struct CfBlock *add_block(const struct State *st)
...
+    const struct Type *from = &obj->type;
 
-    struct CfVariable *destvar = add_variable(st, resulttype, "$cast");
+    if (same_type(from, to))
+        return obj;
 
-    switch(resulttype->kind) {
-    case TYPE_SIGNED_INTEGER:
+    // Casting to bigger signed int applies when "to" is signed and bigger.
+    // Doesn't cast from signed to same size unsigned: with 8 bits, 255 does not implicitly cast to -1.
```
Akuli, line 95:
```suggestion
    // Doesn't cast from unsigned to same size signed: with 8 bits, 255 does not implicitly cast to -1.
```

src/build_cfg.c
```diff
@@ -36,61 +50,213 @@ static struct CfBlock *add_block(const struct State *st)
...
     struct CfVariable *result, *temp;
 
     switch(expr->kind) {
     case AST_EXPR_CALL:
-        result = build_cfg_for_call(st, &expr->data.call, expr->type_before_implicit_cast.kind == TYPE_UNKNOWN ? NULL : &expr->type_before_implicit_cast);
+        result = build_call(st, &expr->data.call, expr->location);
+        if (result == NULL) {
+            if (!needvalue)
+                return NULL;
+            // TODO: add a test for this
```
Akuli, line 245: Should do

src/build_cfg.c
```diff
@@ -137,15 +303,37 @@ static struct CfVariable *build_cfg_for_expression(const struct State *st, const
...
+                // TODO: is this evaluation order good?
+                struct CfVariable *target = build_address_of_expression(st, targetexpr);
+                assert(target->type.kind == TYPE_POINTER);
+                const char *errmsg;
+                switch(targetexpr->kind) {
+                    case AST_EXPR_GET_VARIABLE: errmsg = "cannot assign a value of type FROM to variable of type TO"; break;
+                    case AST_EXPR_DEREFERENCE: errmsg = "cannot assign a value of type FROM into a pointer of type TO*"; break;
+                    default: assert(0);
+                }
+                // TODO: i think "bytevar = (intvar = 'x')" doesn't work, call it bug or feature? (compare to c)
```
Akuli, line 329: add test?

src/jou_compiler.h
```diff
@@ -232,7 +220,7 @@ struct CfInstruction {
         // TODO: replace nargs with NULL terminated?
         struct { char funcname[100]; struct CfVariable **args; int nargs; } call; // CF_CALL
     } data;
-    const struct CfVariable *destvar;  // NULL when it doesn't make sense, e.g. functions that return void
+    struct CfVariable *destvar;  // NULL when it doesn't make sense, e.g. functions that return void
```
Akuli, line 223: Maybe these should be consted instead

Issue #15, opened 2022-12-31
No unreachable warning for some infinite loops


Akuli:
I'm preparing a pull request to show warnings for unreachable code, but it doesn't work in at least these two cases I tried:

```python
def after_infinite_loop_with_variable() -> void:
    flag = True
    while flag:
        puts("hi")
    puts("yooooo wat")  # Warning: this code will never run

def after_infinite_loop_with_variable_set_after_loop() -> void:
    flag = True
    while flag:
        puts("hi")
    flag = False  # Warning: this code will never run
```

Akuli:
Now the first example works as expected. The second one is still broken.

Akuli:
Tried it today, getting no unreachable warnings for this. The problem is that UVGs have a `set flag` instruction that doesn't specify that `flag` is being set to a known-to-be-true value.

<details><summary>The UVG's</summary>

```
$ jou --uvg-only a.jou 
===== UVG for after_infinite_loop_with_variable() -> None =====
block 0 (start):
    [line 5]    statement
    [line 5]    set flag
    [line 6]    statement
    Jump to block 1.
block 1:
    [line 6]    use flag
    Jump to either block 2 or 4 depending on some condition.
block 2:
    [line 7]    statement
    Jump to block 3.
block 3:
    Jump to block 1.
block 4:
    [line 8]    statement
    Return from function.

===== UVG for after_infinite_loop_with_variable_set_after_loop() -> None =====
block 0 (start):
    [line 12]   statement
    [line 12]   set flag
    [line 13]   statement
    Jump to block 1.
block 1:
    [line 13]   use flag
    Jump to either block 2 or 4 depending on some condition.
block 2:
    [line 14]   statement
    Jump to block 3.
block 3:
    Jump to block 1.
block 4:
    [line 15]   statement
    [line 15]   set flag
    Return from function.
```

</details>

PR #17, opened 2023-01-01
Better analyzing for values of booleans in simplify_cfg

Akuli:
Fixes half of #15. The other half requires something to get rid of unnecessary address-ofs: if a variable named `flag` already exists, then `flag = False` becomes `*(&flag) = False` in the CFG, and `simplify_cfg` avoids touching variables whose address is used somewhere.

### DIFF
--- a/src/simplify_cfg.c
+++ b/src/simplify_cfg.c
@@ -21,26 +21,55 @@ struct BoolStatus {
     bool can_be_false;
 };
 
-bool update(struct BoolStatus *dest, const struct BoolStatus *src, int n)
+static bool add_possibilities(struct BoolStatus *dest, const struct BoolStatus *src, int n)
 {
-    bool changed = false;
+    bool did_something = false;
     for (int i = 0; i < n; i++) {
         if (src[i].can_be_true && !dest[i].can_be_true) {
-            dest[i].can_be_true = true;
-            changed = true;
+            dest[i].can_be_true = 1;
+            did_something = true;
         }
         if (src[i].can_be_false && !dest[i].can_be_false) {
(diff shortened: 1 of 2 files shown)


src/simplify_cfg.c
```diff
@@ -16,38 +16,174 @@ static int find_var_index(const struct CfGraph *cfg, const struct CfVariable *v)
...
+            return false;
+    return true;
+}
+
+/*
+Figure out whether boolean variables are true or false.
+return_value[block_index][variable_index] = status of variable at END of block
+variable_index must be so that the variable has type bool.
+
+Idea: Initially mark everything as impossible, so no variable can be true or false
```
Akuli, line 53: Idea has changed a bit since writing this comment.

src/simplify_cfg.c
```diff
@@ -16,38 +16,174 @@ static int find_var_index(const struct CfGraph *cfg, const struct CfVariable *v)
...
+            } else {
+                // What is possible in other blocks is determined based on only how
+                // they are jumped into.
+                tempstatus[i].can_be_true = 0;
+                tempstatus[i].can_be_false = 0;
+            }
+        }
+        for (int i = 0; i < nblocks; i++) {
+            if (cfg->all_blocks.ptr[i]->iftrue == cfg->all_blocks.ptr[visiting]
+                || cfg->all_blocks.ptr[i]->iffalse == cfg->all_blocks.ptr[visiting])
```
Akuli, line 95:
```suggestion
            if (cfg->all_blocks.ptr[i]->iftrue == visitingblock
                || cfg->all_blocks.ptr[i]->iffalse == visitingblock)
```

PR #21, opened 2023-01-01
Rewrite the code that shows unreachable warnings

### LINKED ISSUE #18: annoyingly many unreachability warnings
code

```python3
cdecl printf(format: byte*, ...) -> int

def main() -> int:
    return 0

    i = 0
    while True:
        printf("%d->", i)
        if i == 3:
            printf("end\n")
            break
            printf("yooooooo\n")
        i = i+1
        printf("%d ", i)
```

expected result:

```
compiler warning for file "../asd.jou", line 6: this code will never run
```

actual result:

```
compiler warning for file "../asd.jou", line 12: this code will never run
compiler warning for file "../asd.jou", line 13: this code will never run
compiler warning for file "../asd.jou", line 10: this code will never run
compiler warning for file "../asd.jou", line 8: this code will never run
compiler warning for file "../asd.jou", line 7: this code will never run
compiler warning for file "../asd.jou", line 6: this code will never run
```

Akuli:
It was a mess that just happened to work for test cases. It's still a mess, but at least I know why it works now.

Also, fixes #18

### DIFF
--- a/src/simplify_cfg.c
+++ b/src/simplify_cfg.c
@@ -1,4 +1,5 @@
 #include "jou_compiler.h"
+#include <limits.h>
 
 static int find_block_index(const struct CfGraph *cfg, const struct CfBlock *b)
 {
@@ -161,7 +162,7 @@ static void dump_known_bool_values(const struct CfGraph *cfg, struct BoolStatus 
 }
 #endif
 
-static void clean_branches_where_condition_always_true_or_always_false(struct CfGraph *cfg, bool *did_something)
+static void clean_jumps_where_condition_always_true_or_always_false(struct CfGraph *cfg)
 {
     struct BoolStatus **statuses = determine_known_bool_values(cfg);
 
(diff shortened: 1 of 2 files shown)


src/simplify_cfg.c
```diff
@@ -272,43 +349,31 @@ static void mark_analyzable_variables(struct CfGraph *cfg, bool *did_something)
...
-            nothing_happened_count = 0;
-        else
-            nothing_happened_count++;
-    }
+    do {
+        clean_jumps_where_condition_always_true_or_always_false(cfg);
+    } while (mark_analyzable_variables(cfg));
 
-    //dump_known_bool_values(cfg, determine_known_bool_values(cfg));
+    remove_unreachable_blocks(cfg);
```
Akuli, line 375: This has a potential problem: if there are occurrences of `&foo` but they are all in unreachable blocks, then the variable `foo` will never be marked analyzable. In fact, this is the situation in the second function of issue #15.

PR #27, opened 2023-01-05
Add a warning for attempting to use undefined variables

### LINKED ISSUE #26: Unused variable warning
Should get warning, if:
- variable set but not used
- variable never set

Example:
```python3
declare puts(message: byte*) -> int

def set_but_not_used() -> void:
    message = "hi"
    puts("hi")

# Uses hypothetical `x: int` declaration syntax that doesn't exist yet
def never_set() -> void:
    x: int
    puts("hi")
```

Akuli:
~Fixes #26~ Adds warnings for code where you use or may use a variable before it is assigned.

### DIFF
--- a/src/simplify_cfg.c
+++ b/src/simplify_cfg.c
@@ -26,7 +26,7 @@ enum VarStatus {
     VS_UNDEFINED,  // No value has been set to the variable. Holds a garbage value.
     VS_UNPREDICTABLE,  // Address of variable (&foo) has been used. Give up analzing it.
     /*
-    Longer explanation of VS_UNPREDICTABLE: The value of an unpredictable variable can
+    Longer description of VS_UNPREDICTABLE: The value of an unpredictable variable can
     change in lines of code that seem to have nothing to do with the variable. For
     example, a function that doesn't take &foo as an argument can change the value of foo,
     because the pointer &foo can be stored in some place that the function can access.
--- a/src/simplify_cfg.c
+++ b/src/simplify_cfg.c
@@ -437,11 +437,48 @@ static void remove_unused_variables(struct CfGraph *cfg)
     free(used);
 }
 
+static void warn_about_undefined_variables(struct CfGraph *cfg)
+{
+    enum VarStatus **statuses = determine_var_statuses(cfg);
+
+    for (int blockidx = 0; blockidx < cfg->all_blocks.len; blockidx++) {
+        const struct CfBlock *b = cfg->all_blocks.ptr[blockidx];
+        enum VarStatus *status = statuses[blockidx];
+        for (struct CfInstruction *ins = b->instructions.ptr; ins < End(b->instructions); ins++) {
+            for (int i = 0; i < ins->noperands; i++) {
+                switch(status[find_var_index(cfg, ins->operands[i])]) {
+                case VS_UNVISITED:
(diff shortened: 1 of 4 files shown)


src/simplify_cfg.c
```diff
@@ -17,88 +17,138 @@ static int find_var_index(const struct CfGraph *cfg, const struct CfVariable *v)
...
+        return;
+
+    int destidx = find_var_index(cfg, ins->destvar);
+    if (statuses[destidx] == VS_UNPREDICTABLE)
+        return;
+
+    switch(ins->kind) {
+    case CF_VARCPY:
+        statuses[destidx] = statuses[find_var_index(cfg, ins->operands[0])];
+        assert(statuses[destidx] != VS_UNVISITED);
```
Akuli, line 98: Presumably we should assert this for all variables.

src/simplify_cfg.c
```diff
@@ -17,88 +17,138 @@ static int find_var_index(const struct CfGraph *cfg, const struct CfVariable *v)
...
+        break;
+    case CF_TRUE:
+        statuses[destidx] = VS_TRUE;
+        break;
+    case CF_FALSE:
+        statuses[destidx] = VS_FALSE;
+        break;
+    default:
+        statuses[destidx] = VS_DEFINED;
+        break;
```
Akuli, line 113: förmätting nit

```suggestion
    case CF_TRUE: statuses[destidx] = VS_TRUE; break;
    case CF_FALSE: statuses[destidx] = VS_FALSE; break;
    default: statuses[destidx] = VS_DEFINED; break;
```

src/simplify_cfg.c
```diff
@@ -17,88 +17,138 @@ static int find_var_index(const struct CfGraph *cfg, const struct CfVariable *v)
...
+    }
+}
 
-#if 0   // change to 1 to debug
+#define DebugPrint 0  // change to 1 to see debug prints
```
Akuli, line 117:
```suggestion
#define DebugPrint 0  // change to 1 to see VarStatuses
```

src/simplify_cfg.c
```diff
@@ -400,11 +437,48 @@ static void remove_unused_variables(struct CfGraph *cfg)
...
+                switch(status[find_var_index(cfg, ins->operands[i])]) {
+                case VS_UNVISITED:
+                    assert(0);
+                case VS_TRUE:
+                case VS_FALSE:
+                case VS_DEFINED:
+                case VS_UNPREDICTABLE:
+                    break;
+                case VS_POSSIBLY_UNDEFINED:
+                    assert(ins->operands[i]->name[0] != '$');
```
Akuli, line 458:
```suggestion
                    // The compiler internally creates variables named $foo.
                    // I think they are never used undefined if the compiler works correctly.
                    assert(ins->operands[i]->name[0] != '$');
```

tests/should_succeed/undefined_value_warning.jou
```diff
@@ -0,0 +1,14 @@
+declare puts(s: byte*) -> int
+
+def maybe_uninitialized(n: int) -> void:
+    for i = 0; i < n; i = i+1:
```
Akuli, line 4:
```suggestion
    for i = 0; i < n; i++:
```

PR #35, opened 2023-01-07
And,or,not

Akuli:


src/build_cfg.c
```diff
@@ -386,6 +391,17 @@ static const struct CfVariable *build_expression(
...
             break;
         }
+    case AST_EXPR_AND:
+        result = build_and_or(st, &expr->data.operands[0], &expr->data.operands[1], AND);
+        break;
+    case AST_EXPR_OR:
+        result = build_and_or(st, &expr->data.operands[0], &expr->data.operands[1], OR);
+        break;
+    case AST_EXPR_NOT:
+        temp = build_expression(st, &expr->data.operands[0], &boolType, "value after 'not' must be a boolean, not FROM", true);
```
Akuli, line 401: Need test

src/build_cfg.c
```diff
@@ -432,7 +448,78 @@ static const struct CfVariable *build_expression(
...
+
+    or:
+        # lhs returning True means we don't evaluate rhs
+        if lhs:
+            result = True
+        else:
+            result = rhs
+    */
+    char errormsg[100];
+    sprintf(errormsg, "'%s' only works with booleans, not FROM", andor==AND ? "and" : "or");
```
Akuli, line 475: Need test. for example `if foo == 1 or 2:`

src/build_cfg.c
```diff
@@ -432,7 +448,78 @@ static const struct CfVariable *build_expression(
...
+
+    switch(andor) {
+    case AND:
+        // result = rhs
+        rhs = build_expression(st, rhsexpr, &boolType, errormsg, true);
+        add_unary_op(st, rhsexpr->location, CF_VARCPY, rhs, result);
+        break;
+    case OR:
+        // result = True
+        ins = add_constant(st, lhsexpr->location, ((struct Constant){.type=boolType,.value.boolean=true}), result);
```
Akuli, line 497: `((struct Constant){...})` belongs in a macro maybe

tests/should_succeed/and_or_not.jou
```diff
@@ -0,0 +1,95 @@
...
+    printf("%d", 1+2==4 or 1+2==4)
+    putchar('\n')
+
+    # Output: 01
+    printf("%d", not 1+2==3)
+    printf("%d", not 1+2==4)
+    putchar('\n')
+
+    # Anything with "False and ..." contains code that never runs: the "..." part
+    # is not evaluated at all because the result will be False anyway.
```
Akuli, line 29:
```suggestion
    # is not evaluated at all because the result will be False anyway.
    #
    # Technically "True and ..." also contains code that never runs: the left side
    # of "and" is never False, but the compiler still generates code to handle that
    # case (the generated code sets the result of the "and" operation to False).
    # Showing a warning about the generated code would probably be more
    # confusing than helpful, so the compiler hides it.
    #
    # Similar things apply to "or".
```

src/build_cfg.c
Akuli: decided not

PR #36, opened 2023-01-09
void pointers

Akuli:


src/types.c
```diff
@@ -36,16 +37,22 @@ struct Type create_integer_type(int size_in_bits, bool is_signed)
...
     case TYPE_POINTER:
         {
             struct Type t2 = *t;
-            t2.data.valuetype = malloc(sizeof(*t2.data.valuetype));
-            *t2.data.valuetype = copy_type(t->data.valuetype);
+            if (t->data.valuetype) {
+                // not a void pointer
+                t2.data.valuetype = malloc(sizeof(*t2.data.valuetype));
+                *t2.data.valuetype = copy_type(t->data.valuetype);
+            }
```
Akuli, line 55: changes like this one should be revered, in this file

PR #38, opened 2023-01-10
Introduce AstType

Akuli:
Currently AST contains `Type`s, but that won't be good once we have structs: the AST building functions don't have any state apart from the stream of tokens (and I don't want to add any), so can't store a list of previously defined structs there.

This changes the AST to have types in a not-fully-compiled form so that the next compilation step (build_cfg) can handle struct names later.

Purely refactoring / adding boilerplate. No functionality changes.

### DIFF
--- a/src/print.c
+++ b/src/print.c
@@ -94,11 +94,26 @@ void print_tokens(const Token *tokens)
     printf("\n");
 }
 
-static void print_ast_function_signature(const Signature *sig, int indent)
+static void print_ast_type(const struct AstType *t)
 {
-    char *s = signature_to_string(sig, true);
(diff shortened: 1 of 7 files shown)


src/print.c
```diff
@@ -94,11 +94,26 @@ void print_tokens(const Token *tokens)
     printf("\n");
 }
 
-static void print_ast_function_signature(const Signature *sig, int indent)
+static void print_ast_type(const struct AstType *t)
```
Akuli, line 97: I keep writing `struct`. Should probably get rid of the typedefs, they also get in the way slightly when doing jump-to-definition.

PR #54, opened 2023-01-11
Add ++ and -- operators for pointers

Akuli:
But in general, cannot do ptr+int or ptr-int yet.

### DIFF
--- a/tests/should_succeed/pointer.jou
+++ b/tests/should_succeed/pointer.jou
@@ -11,6 +11,10 @@ def myputs1(s: byte*) -> void:
     for i = 0; s[i] != '\0'; i++:
         putchar(s[i])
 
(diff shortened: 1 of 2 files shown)


tests/should_succeed/pointer.jou
```diff
@@ -11,12 +11,17 @@ def myputs1(s: byte*) -> void:
...
+        putchar(*s++)
+
 def main() -> int:
     # Output: hhii
     putchar2('h')
     putchar2('i')
     putchar('\n')
 
     myputs1("asd\n")  # Output: asd
+    myputs1("foo\n")  # Output: foo
```
Akuli, line 25:
```suggestion
    myputs2("foo\n")  # Output: foo
```

PR #61, opened 2023-01-14
Separate type checking and build_cfg

Akuli:
Adds quite a lot more code, but will be needed for arrays: `foo[bar]` needs to take the address of the `foo` variable if it's an array, and otherwise it uses the *value* of the variable as the pointer.

### DIFF
--- a/src/build_cfg.c
+++ b/src/build_cfg.c
@@ -707,36 +707,6 @@ static CfGraph *build_function(struct State *st, const AstBody *body)
     return st->cfg;
 }
 
-static Signature build_signature(const struct State *st, const AstSignature *astsig, Location location)
-{
-    for (int i = 0; i < st->typectx.function_signatures.len; i++)
-        if (!strcmp(st->typectx.function_signatures.ptr[i].funcname, astsig->funcname))
-            fail_with_error(location, "a function named '%s' already exists", astsig->funcname);
-
-    Signature result = { .nargs = astsig->nargs, .takes_varargs = astsig->takes_varargs };
-    safe_strcpy(result.funcname, astsig->funcname);
-
-    size_t size = sizeof(result.argnames[0]) * result.nargs;
-    result.argnames = malloc(size);
-    memcpy(result.argnames, astsig->argnames, size);
-
-    result.argtypes = malloc(sizeof(result.argtypes[0]) * result.nargs);
-    for (int i = 0; i < result.nargs; i++)
-        result.argtypes[i] = type_from_ast(&st->typectx, &astsig->argtypes[i]);
-
-    result.returntype = type_or_void_from_ast(&st->typectx, &astsig->returntype);
-    // TODO: validate main() parameters
-    // TODO: test main() taking parameters
-    if (!strcmp(astsig->funcname, "main") &&
-        (result.returntype == NULL || !same_type(result.returntype, &intType)))
-    {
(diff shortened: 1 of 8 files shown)


src/build_cfg.c
```diff
@@ -2,46 +2,44 @@
...
+}
+
 static const Signature *find_function(const struct State *st, const char *name)
 {
-    for (int i = 0; i < st->cfgfile->nfuncs; i++)
-        if (!strcmp(st->cfgfile->signatures[i].funcname, name))
-            return &st->cfgfile->signatures[i];
+    for (int i = 0; i < st->typectx.function_signatures.len; i++)
+        if (!strcmp(st->typectx.function_signatures.ptr[i].funcname, name))
+            return &st->typectx.function_signatures.ptr[i];
```
Akuli, line 42: - Could loop with pointers instead of indexes
- Do we really need this in `build_cfg`?

src/build_cfg.c
```diff
@@ -1064,13 +708,25 @@ static CfGraph *build_function(struct State *st, const Signature *sig, const Ast
...
+
+    for (ExpressionTypes **et = st->typectx.expr_types.ptr; et < End(st->typectx.expr_types); et++) {
+        free_type(&(*et)->type);
+        if ((*et)->type_after_cast) {
+            free_type((*et)->type_after_cast);
+            free((*et)->type_after_cast);
+        }
+        free(*et);
+    }
+    st->typectx.expr_types.len = 0;
```
Akuli, line 722: This mess should go into a new function

src/build_cfg.c
```diff
@@ -1064,13 +708,25 @@ static CfGraph *build_function(struct State *st, const Signature *sig, const Ast
...
     return st->cfg;
 }
 
 static Signature build_signature(const struct State *st, const AstSignature *astsig, Location location)
 {
-    for (int i = 0; i < st->cfgfile->nfuncs; i++)
-        if (!strcmp(st->cfgfile->signatures[i].funcname, astsig->funcname))
+    for (int i = 0; i < st->typectx.function_signatures.len; i++)
+        if (!strcmp(st->typectx.function_signatures.ptr[i].funcname, astsig->funcname))
```
Akuli, line 729: Could rewrite this loop with pointers

src/jou_compiler.h
```diff
@@ -315,14 +317,34 @@ char *signature_to_string(const Signature *sig, bool include_return_type);
...
+    const Signature *current_function_signature;
+    // expr_types tells what type each expression has.
+    // It contains nothing for calls to "-> void" functions.
+    List(ExpressionTypes *) expr_types;
+    List(Variable *) variables;
+    List(Type) structs;
+    List(Signature) function_signatures;
+};
+Type type_from_ast(const TypeContext *ctx, const AstType *asttype);
+Type *type_or_void_from_ast(const TypeContext *ctx, const AstType *asttype);
```
Akuli, line 342: Maybe there should also be a `signature_from_ast` function. Currently that is done in build_cfg

src/typecheck.c
```diff
@@ -0,0 +1,665 @@
+#include "jou_compiler.h"
+
+
+// If error_location is NULL, this will return NULL when variable is not found.
```
Akuli, line 4:
```suggestion
```

PR #66, opened 2023-01-15
JIT

Akuli:
Todo:
- [x] remove `--no-jit` option and clang code (turns out JIT is as fast as clang, at least for a primitive fib benchmark)
- [ ] print optimized LLVM IR, would be curious to see it... :)
- [x] run tests with different optimization levels (except tests that are intentionally a bit UB)
- [x] fix the `--help` option
- [ ] refactor / cleanup

Maybe I will split this PR into two: one PR to add the JIT and another to add command-line optimization options and such.

### DIFF
--- a/src/run.c
+++ b/src/run.c
@@ -1,0 +1,105 @@
+#include <errno.h>
+#include <string.h>
+#include "jou_compiler.h"
+#include <llvm-c/Core.h>
+#include <llvm-c/ExecutionEngine.h>
+
+// TODO: this TempDir code is ridiculous
+static char TempDir[50];
+
+static void delete_temp_dir()
+{
+    char command[200];
+    sprintf(command, "rm -rf '%s'", TempDir);
+    system(command);
+}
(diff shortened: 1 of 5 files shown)


src/run.c
```diff
@@ -1,60 +1,64 @@
...
 
-static char TempDir[50];
+    LLVMPassManagerRef pm = LLVMCreatePassManager();
 
-static void cleanup()
-{
-    char command[200];
-    sprintf(command, "rm -rf '%s'", TempDir);
-    system(command);
+    LLVMPassManagerBuilderRef pmbuilder = LLVMPassManagerBuilderCreate();
```
Akuli, line 11:
```suggestion
    /*
    The default settings should be fine for Jou because they work well for
    C and C++, and Jou is quite similar to C.
    */
    LLVMPassManagerBuilderRef pmbuilder = LLVMPassManagerBuilderCreate();
```

PR #67, opened 2023-01-15
Add -O0,-O1,-O2,-O3 command line arguments + related refactoring

Akuli:


src/jou_compiler.h
```diff
@@ -443,6 +451,7 @@ AstToplevelNode *parse(const Token *tokens);
 CfGraphFile build_control_flow_graphs(AstToplevelNode *ast);
 void simplify_control_flow_graphs(const CfGraphFile *cfgfile);
 LLVMModuleRef codegen(const CfGraphFile *cfgfile);
+int run_program(LLVMModuleRef module, const CommandLineFlags *flags);  // destroys the module
```
Akuli, line 454:
```suggestion
int run_program(LLVMModuleRef module, const CommandLineFlags *flags);
```

PR #69, opened 2023-01-15
add hex and bin numbers

Akuli:
Also fixed tokenizing to support 64 bit values.

### DIFF
--- a/src/jou_compiler.h
+++ b/src/jou_compiler.h
@@ -75,7 +75,7 @@ struct Token {
     } type;
     Location location;
     union {
-        int int_value;  // TOKEN_INT
+        long long int_value;  // TOKEN_INT
         char char_value;  // TOKEN_CHAR
         char *string_value;  // TOKEN_STRING
         int indentation_level;  // TOKEN_NEWLINE, indicates how many spaces after newline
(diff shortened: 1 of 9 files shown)


Akuli:
Needs a better error for attempting to use a variable name like `2bad`

src/jou_compiler.h
```diff
@@ -75,7 +75,7 @@ struct Token {
     } type;
     Location location;
     union {
-        int int_value;  // TOKEN_INT
+        long long int_value;  // TOKEN_INT
```
Akuli, line 78: Presumably this could be `unsigned long long`, as it is meant to hold 64 bits unsigned.

tests/syntax_error/unnecessary_zero.jou
```diff
@@ -0,0 +1,13 @@
...
+    # In C this would be 777 in octal (base 8).
+    # IMO it is more confusing than helpful.
+    #
+    # Leading zeros are allowed in hex and binary numbers, because in those
+    # numbers, the position is usually significant. Example:
+    #
+    #   FLAG_FOO = 0x01
+    #   FLAG_BAR = 0x02
+    #   FLAG_BAZ = 0x04
+    #   FLAG_LOL = 0x10
```
Akuli, line 11:
```suggestion
    #   FLAG_LOL = 0x08
    #   FLAG_WAT = 0x10
```

src/jou_compiler.h
Akuli: maybe later

PR #75, opened 2023-01-17
Add documentation about performance and optimizations

Akuli:


Akuli:
Note to myself: I will need to change the examples to use `import`.

PR #79, opened 2023-01-18
Fix accessing fields of non-"addressable" struct objects

### LINKED ISSUE #40: confusing & error with temporary struct object
I will soon create a PR that adds structs. It will include a funny corner case...

```python3
struct Foo:
    x: int

def foo() -> void:
    lol = Foo{x=1}.x
```

Expected result: no error or an understandable error.

Actual result: `compiler error in file "asd.jou", line 5: the address-of operator '&' cannot be used with a newly calculated value`

The error message is confusing because it talks about `&` operator which doesn't even appear in the code.

This happens because to evaluate `foo.bar`, we need to compute the offset of `bar` in the struct and add it to the address of `foo`. So we need to know the address of `foo`. In this case it means finding the address (&) of a temporary value which is not allowed.

Akuli:
Fixes #40

### DIFF
--- a/src/build_cfg.c
+++ b/src/build_cfg.c
@@ -177,8 +177,19 @@ static const Variable *build_struct_field_pointer(
     assert(0);
 }
 
+static const Variable *build_struct_field(
+    struct State *st, const Variable *structinstance, const char *fieldname, Location location)
+{
+    const Variable *ptr = add_variable(st, get_pointer_type(structinstance->type));
+    add_unary_op(st, location, CF_ADDRESS_OF_VARIABLE, structinstance, ptr);
+    const Variable *field_ptr = build_struct_field_pointer(st, ptr, fieldname, location);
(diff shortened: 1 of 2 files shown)


src/build_cfg.c
```diff
@@ -398,6 +409,9 @@ static const Variable *build_expression(struct State *st, const AstExpression *e
         result = build_struct_init(st, types->type, &expr->data.call, expr->location);
         break;
     case AST_EXPR_GET_FIELD:
+        temp = build_expression(st, expr->data.field.obj);
+        result = build_struct_field(st, temp, expr->data.field.fieldname, expr->location);
+        break;
     case AST_EXPR_DEREF_AND_GET_FIELD:
     case AST_EXPR_INDEXING:
         // To evaluate foo->bar, we first evaluate &foo->bar and then dereference.
```
Akuli, line 417: This comment needs to be expanded a bit.

PR #96, opened 2023-01-19
Import improvements

### LINKED ISSUE #82: Import brings in all symbols instead of only what was asked
This shouldn't work, but it does:

```python
from "stdlib/io.jou" import scanf

def main() -> int:
    puts("Hi")
    return 0
```

### LINKED ISSUE #83: Importing a symbol that doesn't exist
there should be error for this:

```python3
from "stdlib/io.jou" import asdasdasdasd
```

Akuli:
Fixes #82 
Fixes #83

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -74,6 +74,7 @@ usage:
 struct FileState {
     char *filename;
     AstToplevelNode *ast;
+    TypeContext typectx;
     LLVMModuleRef module;
 };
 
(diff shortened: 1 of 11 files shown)


src/main.c
```diff
@@ -155,17 +130,48 @@ static void parse_file(struct CompileState *compst, const char *filename, const
...
+
+        const Signature *sig = NULL;
+        for (Signature *p = src->typectx.exports.ptr; p < End(src->typectx.exports); p++) {
+            if (!strcmp(p->funcname, impnode->data.import.symbol)) {
+                sig = p;
+                break;
+            }
+        }
+        if (!sig) {
+            // TODO: test
```
Akuli, line 165: Already tested.
```suggestion
```

PR #102, opened 2023-01-20
prefix - operator

### LINKED ISSUE #101: -1
Currently the only way to create a negative number is to write e.g. `0-1`. I think a simple `-1` would be better.

It should be parsed probably like this: whenever we are parsing an expression with additions and subtractions, we check if the first token is `-`, and in that case the first term is negated. So `-x+y` parses as `Add(Neg(Var(x)), Var(y))`.

Akuli:
Fixes #101

### DIFF
--- a/src/build_cfg.c
+++ b/src/build_cfg.c
@@ -88,7 +88,7 @@ static CfInstruction *add_instruction(
 #define add_binary_op(st, loc, op, lhs, rhs, target) \
     add_instruction((st), (loc), (op), NULL, (const Variable*[]){(lhs),(rhs),NULL}, (target))
 #define add_constant(st, loc, c, target) \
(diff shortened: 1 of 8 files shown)


src/build_cfg.c
```diff
@@ -393,6 +393,22 @@ static const Variable *build_struct_init(struct State *st, const Type *type, con
...
+{
+    assert(is_integer_type(t));
+    Constant c = {
+        .kind = CONSTANT_INTEGER,
+        .data.integer = {
+            .width_in_bits = t->data.width_in_bits,
+            .is_signed = (t->kind == TYPE_SIGNED_INTEGER),
+            .value = 0,
+        },
+    };
```
Akuli, line 406: This kind of code appears in many places. Maybe it's time to make it a function.

PR #103, opened 2023-01-21
codeblocks config and notes.txt

### LINKED ISSUE #78: Windows Support
Really excited to try this

Akuli:
Fixes #78 

Still needs work:
- Figure out the best way to get the llvm-c headers
- Rewrite winnotes.txt into somewhat reasonable instructions
- Set up GitHub Actions

The llvm-c headers are weird. On Windows, `C:\Program Files\LLVM\include\llvm-c` contains only two files, `lto.h` and `Remarks.h`. It is supposed to contain much more files, e.g. on my linux system `/usr/include/llvm-c-13/llvm-c/` contains 26 headers and 1 subdirectory.

For now I copied it all to a directory named `hax\llvm-c` within the Jou project, and because those headers included other headers, I also copied `/usr/include/llvm-13/llvm/` to `hax\llvm`. It is a **lot** of headers, no way I'm adding it all to git. Maybe a script to download it all?


jou.cbp
```diff
@@ -0,0 +1,85 @@
...
+				<Option output="./jou" prefix_auto="1" extension_auto="1" />
+				<Option object_output="obj/Debug/" />
+				<Option type="1" />
+				<Option compiler="clang" />
+				<Compiler>
+					<Add option="-g" />
+				</Compiler>
+			</Target>
+			<Target title="Release">
+				<Option output="bin/Release/jou" prefix_auto="1" extension_auto="1" />
```
Akuli, line 19: Don't think `bin/` is gitignored

jou.cbp
```diff
@@ -0,0 +1,85 @@
...
+					<Add option="-s" />
+				</Linker>
+			</Target>
+		</Build>
+		<Compiler>
+			<Add option="-Wall" />
+			<Add option="-std=c11" />
+			<Add option="-target x86_64-w64-windows-gnu" />
+			<Add directory="C:/Program Files/LLVM/include" />
+			<Add directory="hax" />
```
Akuli: Will need to remove the `hax` directory when I figure out a better way.

sumeshir26:
Instead of installing LLVM manually, it seems like it is easily possible to install clang and LVMM through Visual Studio: https://stackoverflow.com/questions/63914108/using-clang-in-windows-10-for-c-c/63916235#63916235:~:text=4-,Expanding,-on%20HolyBlackCat%27s%20answer

sumeshir26:
I managed to install clang, but this is what I get:
image

Akuli:
I'm not going to use Visual Studio or anything else that is created by microsoft and not open-source, except Windows because I kinda need to :) This is the main reason why I chose codeblocks.

Because I already got things to work with codeblocks, it would be easiest for you to use codeblocks as well. After all, it is open-source, and IMO there's not much to dislike.

That said, you can use visual studio if you really really want to. To do so you need to figure out:
- Did a `Core.h` file actually get installed inside a directory named `llvm-c` somewhere? If it did, you need to pass `"-IC:\Program Files\BlaBlaBla` argument to clang, where `C:\Program Files\BlaBlaBla` is the directory containing `llvm-c`. If not, download `llvm_headers.zip` from this PR and extract it.
- Did a `LLVM-C.lib` file get installed somewhere? You need to pass it to the compiler when linking.
- How to properly compile Jou. You don't run `clang` on the `.h` files. You instead run it on each `.c` file to compile them into object files, and then one more time on all object files to link them into an `.exe` file. It would be great if your IDE did it for you (like codeblocks does).

Akuli:
Can you try with the instructions in this PR? I tried to make them very easy to follow.

sumeshir26:
> Can you try with the instructions in this PR? I tried to make them very easy to follow.

OKay, will try tommorow

Akuli:
I will now merge this PR, so that I can work on other stuff that is based on this PR. This also means that you can find the instructions in the README at https://github.com/Akuli/jou

Akuli:
@sumeshir26 I figured out an even easier way to run Jou on windows: now github actions creates a `.zip` file that contains everything you need already compiled, `jou.exe` and a bunch of `.dll` files. You can basically download the zip file and you're almost ready to go.

The updated instructions are in README (scroll down to "Setup").

sumeshir26:
> @sumeshir26 I figured out an even easier way to run Jou on windows: now github actions creates a `.zip` file that contains everything you need already compiled, `jou.exe` and a bunch of `.dll` files. You can basically download the zip file and you're almost ready to go.
> 
> The updated instructions are in README (scroll down to "Setup").

Interesting, but the same process needs do be done every time the repo updates.
Maybe I will try to write a PowerShell script which automates it locally, so I can use git to update easily.

Akuli:
That's pretty easy with GitHub's [rest api](https://docs.github.com/en/rest/releases/assets?apiVersion=2022-11-28#list-release-assets). First you need to get the ID of the latest release by parsing JSON from https://api.github.com/repos/Akuli/jou/releases/latest. The first few lines are:

```
$ curl -s https://api.github.com/repos/Akuli/jou/releases/latest | head
{
  "url": "https://api.github.com/repos/Akuli/jou/releases/89818869",
  "assets_url": "https://api.github.com/repos/Akuli/jou/releases/89818869/assets",
  "upload_url": "https://uploads.github.com/repos/Akuli/jou/releases/89818869/assets{?name,label}",
  "html_url": "https://github.com/Akuli/jou/releases/tag/2023-01-22",
  "id": 89818869,
  "author": {
    "login": "github-actions[bot]",
    "id": 41898282,
    "node_id": "MDM6Qm90NDE4OTgyODI=",
```

Here `89818869` is the release ID. Once you have that, you can ask what assets that release has and get a download URL:

```
$ curl -s https://api.github.com/repos/Akuli/jou/releases/89818869/assets | grep browser_download_url
    "browser_download_url": "https://github.com/Akuli/jou/releases/download/2023-01-22/jou_windows_64bit_2023-01-22.zip"
```

Maybe I should add a `jou --update` command that does this and then downloads the resulting zip file.

Akuli:
I started implementing `jou --update`, but I realized it would be a pain without a good package manager: I need a library to download a file over HTTPS, a second library to parse JSON, and a third library to extract a zip file.

I think the best solution would be to create a powershell script that does the updating (I'll let you do that), and then make it so that `jou --update` runs that powershell script.

sumeshir26:
> 

Sure, I will get started on the PowerShell script soon

PR #119, opened 2023-01-22
Do not rely on argv[0] to locate the executable

### LINKED ISSUE #118: stdlib not found when it isn't in current directory
Hello world program works only if I change directory to `Desktop\jou`, which is where I downloaded and extracted Jou.

```
C:\Users\a\Desktop>jou hello.jou
compiler error in file "hello.jou", line 1: cannot import from "stdlib/io.jou":
No such file or directory

C:\Users\a\Desktop>cd jou

C:\Users\a\Desktop\jou>jou ..\hello.jou
Hello World

C:\Users\a\Desktop\jou>
```

Akuli:
Fixes #118

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -248,15 +248,9 @@ static char *find_stdlib()
     char *exe = find_this_executable();
     const char *exedir = dirname(exe);
 
-    char *path;
-    if (!strcmp(exedir, ".")) {
-        // ./stdlib looks a bit ugly in error messages and debug output IMO
-        path = strdup("stdlib");
-    } else {
-        path = malloc(strlen(exedir) + 10);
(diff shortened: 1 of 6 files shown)


src/main.c
```diff
@@ -202,27 +209,64 @@ static void compile_ast_to_llvm(struct CompileState *compst, struct FileState *f
...
-    strcpy(path, exedir);
-    free(s);
-    strcat(path, "/stdlib");
+static char *find_stdlib()
+{
+    char *exe = find_this_executable();
+    const char *exedir = dirname(exe);
+
+    char *path;
+    if (!strcmp(exedir, ".")) {
```
Akuli, line 252: this never happens

PR #126, opened 2023-01-23
Add Update Script

sumeshir26:
Here is how the script works:
1. Downloads the zip file from GitHub releases
2. Extracts it to `./jou_update`
3. Copies `./jou_update` to `./`
4. Deletes `./jou_update` and `./jou_update.zip`
Hope this is fine, tell me if there are any changes to be made.


Akuli:
This PR now contains also the code to run the script when typing `jou --update`. I'll merge this, so that a new release will appear in a few hours and you can see if it works at all :D

PR #128, opened 2023-01-24
Auto-updater fixes

Akuli:


Akuli:
I ended up downloading Windows 10 for use in a VM. With these changes, the auto-updater actually works.

PR #130, opened 2023-01-24
jou --update: Avoid erasing the installation directory on Windows

### LINKED ISSUE #129: Windows auto-updater script erases the installation directory
https://github.com/Akuli/jou/blob/7ba6a2d62558058504a370a87c057b90fe91a6df/update.ps1#L32-L38

This deletes everything in the Jou installation directory (except `jou_update` which contains the new Jou version, and any files that cannot be deleted). This is:

- good, because any files that are in the old Jou version but not in the new Jou version will be deleted
- bad, because the user might store their files in the installation directory (ideally they wouldn't do that).

@sumeshir26 Be aware of this if you run my version of the updating script! (If this issue is closed by the time you see it, ignore this warning, it means I have already fixed it)

I see a few solutions:
- Delete only `jou.exe`, `*.dll`, `stdlib`. Seems like the best option. If a user modifies `stdlib` or adds their own dll files, they're doing something really weird anyway :)
- Include a list of files with each Jou zip, and delete only the files that are in the list.
- Include a list of deleted files that previously existed in each Jou zip, and delete only those files, overwriting everything else.

Akuli:
Fixes #129

### DIFF
--- a/update.ps1
+++ b/update.ps1
@@ -36,8 +36,7 @@ Remove-Item jou_update.zip
 #
 # We can't erase the whole installation directory because the user might
 # put their files there. It isn't a good idea anyway, but we really
-# shouldn't wipe the user's files. Instead we delete only a few carefully
-# chosen files.
+# shouldn't wipe the user's files.
 Write-Output "Deleting old Jou..."
 Remove-Item stdlib -Recurse
 Rename-Item jou.exe jou.exe.old
--- a/update.ps1
+++ b/update.ps1
@@ -41,11 +41,11 @@ Remove-Item jou_update.zip
 Write-Output "Deleting old Jou..."
 Remove-Item stdlib -Recurse
 Rename-Item jou.exe jou.exe.old
-Get-ChildItem . -Filter *.dll | ForEach-Object {
+foreach ($dll in Get-ChildItem . -Filter *.dll) {
     try {
-        Remove-Item $_
+        Remove-Item $dll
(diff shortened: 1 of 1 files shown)


update.ps1
```diff
@@ -32,11 +32,23 @@ Remove-Item jou_update.zip
 # jou.exe and some dll files are currently running, so they cannot be
 # deleted, overwritten, or moved. But they can be renamed.
 #
-# Most files can still be deleted, so we delete as much as we can.
+# Most files (e.g. most DLLs, all of stdlib) can be deleted as usual.
```
Akuli, line 35:
```suggestion
# Most files (e.g. all of stdlib) can be deleted as usual.
```

update.ps1
```diff
@@ -32,11 +32,23 @@ Remove-Item jou_update.zip
 # jou.exe and some dll files are currently running, so they cannot be
 # deleted, overwritten, or moved. But they can be renamed.
 #
-# Most files can still be deleted, so we delete as much as we can.
+# Most files (e.g. all of stdlib) can be deleted as usual.
+#
+# We can't erase the whole installation directory because the user might
+# put their files there. It isn't a good idea anyway, but we really
+# shouldn't wipe the user's files. Instead we delete only a few carefully
+# chosen files.
```
Akuli, line 40:
```suggestion
# shouldn't wipe the user's files.
```

PR #134, opened 2023-01-24
Combine tests into fewer, longer files

Akuli:
Speeds up valgrinding, because it takes a while for valgrind to start. Also gets rid of some copy/pasta.


Akuli:
Impossible to review. Will split up into smaller PRs.

Akuli:
With many smaller PRs merged, this is no longer impossible to review.

PR #142, opened 2023-01-25
Arrays

### LINKED ISSUE #111: Arrays
Not sure if I want to have an array type, like in C. The other alternative is that a statement `foo = [1,2,3]` allocates stack space for 3 ints, stores the ints there, and then makes `foo` of type `int*` point at the beginning.

Advantages with array type:
- `sizeof(foo)/sizeof(foo[0])` is fun
- determining the length of a hard-coded array is useful when looping over it

Disadvantages with array type:
- limited to fixed length values, unless you risk a stack overflow or you do something more magical than C does
- complicated to implement (need to implicitly cast the array to a pointer to the array's first element, which is essentially a pointer to the array)

Akuli:
Fixes #111


src/typecheck.c
```diff
@@ -32,39 +32,22 @@ static const Signature *find_function(const TypeContext *ctx, const char *name)
...
+        && expr->data.constant.data.integer.is_signed
+        && expr->data.constant.data.integer.width_in_bits == 32)
+    {
+        return (int)expr->data.constant.data.integer.value;
     }
 
-    while (npointers--)
-        t = get_pointer_type(t);
-    return t;
+    fail_with_error(expr->location, "cannot evaluate array length at compile time");
```
Akuli, line 45: should test

src/typecheck.c
```diff
@@ -372,7 +394,10 @@ static const Type *typecheck_indexing(
             indextype->name);
     }
 
-    return ptrtype->data.valuetype;
+    if (ptrtype->kind == TYPE_ARRAY)
+        return ptrtype->data.array.membertype;
+    else
+        return ptrtype->data.valuetype;
```
Akuli, line 400: i don't like this

tests/other_errors/address_of_array_indexing.jou
```diff
@@ -0,0 +1,15 @@
...
+# We can't always evaluate &array, e.g. when it is a function call.
+def foo() -> int[3]:
+    result: int[3]
+    result[0] = 12
+    result[1] = 34
+    result[2] = 56
+    return result
+
+def bar() -> void:
+    x = &foo()[6]  # Error: cannot create a pointer into an array that comes from a function call
```
Akuli, line 15:
```suggestion
    x = &foo()[1]  # Error: cannot create a pointer into an array that comes from a function call
```

tests/should_succeed/array.jou
```diff
@@ -0,0 +1,38 @@
...
+    result[2] = 6
+    return result
+
+# Arrays are passed by value.
+def do_nothing(a: int[3]) -> void:
+    a[0] = a[0] + 7
+
+# Use a pointer instead if you want to modify the array.
+def increment(a: int*) -> void:
+    a[0] = a[0] + 7
```
Akuli, line 18: maybe should be `++a[0]` instead, so it better matches the name

src/typecheck.c
Akuli: I don't like the alternatives any better :)

PR #143, opened 2023-01-25
prefer llvm 11 over 13, because its faster

Akuli:
Makes a noticeable difference when valgrinding.


Akuli:
`time ./runtests.sh --valgrind` before: 1min 45sec

`time ./runtests.sh --valgrind` after: 1min 22sec

Not as huge improvement as I was hoping it would be. Also not great for anyone trying to compile Jou after a few years when LLVM 11 will no longer be commonly available.

PR #145, opened 2023-01-25
Split valgrinding across two github actions jobs

Akuli:
Valgrinding is the slowest thing in the CI. This should speed up a bit.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -10,15 +10,14 @@ jobs:
     strategy:
       matrix:
         llvm-version: [11, 13]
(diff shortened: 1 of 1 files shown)


.github/workflows/linux.yml
```diff
@@ -10,15 +10,14 @@ jobs:
     strategy:
       matrix:
         llvm-version: [11, 13]
+        opt-level: ['-O0', 'O3']
```
Akuli, line 13:
```suggestion
        opt-level: ['-O0', '-O3']
```

Akuli:
Much better now: slowest job is 2min30sec, previously 4min.

PR #171, opened 2023-01-31
Float

littlewhitecloud:
This is float type for "jou".
I haven't test yet, beacause on my computer, It has some (%^#$) error. And can't compile well...
So who can help me test my fork and create issues on my fork. I will fix it and fix the pull request...
Thx


littlewhitecloud:
wait, I founded actions to check

Akuli:
I changed GitHub's settings so that github actions should run when you push.

src/jou_compiler.h
```diff
@@ -106,6 +108,7 @@ struct Constant {
         struct { int width_in_bits; bool is_signed; long long value; } integer;
         char *str;
         char double_text[100];  // convenient because LLVM wants a string anyway
+        char float_text[100]; // same as double
```
Akuli, line 111: Why do we need two of these? Maybe just have one `float_or_double_text` member?

src/jou_compiler.h
```diff
@@ -324,6 +327,7 @@ struct Type {
         TYPE_UNSIGNED_INTEGER,
         TYPE_BOOL,
         TYPE_DOUBLE,
+        TYPE_FLOAT,
```
Akuli, line 330: I think we don't need both `TYPE_DOUBLE` and `TYPE_FLOAT` here, similarly to how `int` and `long` both use `TYPE_SIGNED_INTEGER`. IMO it would be cleaner to distinguish them using `width_in_bits`, like we already do for integers.

src/jou_compiler.h
```diff
@@ -370,7 +375,8 @@ Type *create_struct(
 void free_type(Type *type);
 
 bool is_integer_type(const Type *t);  // includes signed and unsigned
-bool is_number_type(const Type *t);  // integers, floats, doubles
+bool is_number_type(const Type *t);  // integers, doubles
+bool is_float_type(const Type *t); // floats need split with doubles
```
Akuli, line 379: What's going on here? It doesn't make any sense to me. Floats are numbers, so `is_number_type()` should check for integers, floats and doubles, as the original comment said.

src/tokenize.c
```diff
@@ -396,6 +396,8 @@ static Token read_token(struct State *st)
                 if (is_keyword(t.data.name))
                     t.type = TOKEN_KEYWORD;
                 else if ('0'<=t.data.name[0] && t.data.name[0]<='9') {
+                    if (is_vaild_float(t.data.name))
+                        t.type = TOKEN_FLOAT;
```
Akuli, line 400: To detect floats, you could do something similar to what I did with longs, i.e. look for `f` suffix.

Or maybe just get rid of float literals, at least for this PR: `12.34f` would do the same thing as `12.34 as float` anyway, so it's not really necessary.

src/typecheck.c
```diff
@@ -159,7 +160,9 @@ static void do_implicit_cast(
             && !(from->kind == TYPE_SIGNED_INTEGER && to->kind == TYPE_UNSIGNED_INTEGER)
         ) || (
             // Cast from any integer type to double.
-            is_integer_type(from) && to == doubleType
+            is_integer_type(from) && to == doubleType
+        ) || (
+            is_integer_type(from) && to == floatType
```
Akuli, line 165: Assuming `is_float_type()` checks for `float` or `double`:
```suggestion
            is_integer_type(from) && is_float_type(to)
```

src/typecheck.c
```diff
@@ -263,6 +267,9 @@ static const Type *check_binop(
     }
     if (got_pointers) {
         cast_type = voidPtrType;
+    }
+    if (got_numbers && got_floats) {
+        cast_type = floatType;
```
Akuli, line 272: This logic needs more work IMO. Here's how it should behave, I think:
- `int + float = float`
- `int + double = double`
- `float + float = float`
- `double + double = double`
- `float + double = double`

So use `double` whenever one of the two sides is a double.

src/typecheck.c
```diff
@@ -617,10 +624,10 @@ static ExpressionTypes *typecheck_expression(TypeContext *ctx, const AstExpressi
         break;
     case AST_EXPR_NEG:
         result = typecheck_expression(ctx, &expr->data.operands[0])->type;
-        if (result->kind != TYPE_SIGNED_INTEGER && result->kind != TYPE_DOUBLE)
+        if (result->kind != TYPE_SIGNED_INTEGER && result->kind != TYPE_DOUBLE && result->kind != TYPE_FLOAT)
```
Akuli, line 627: Same thing, but simpler:
```suggestion
        if (result->kind != TYPE_SIGNED_INTEGER && !is_float_type(result))
```

src/types.c
```diff
@@ -133,6 +136,11 @@ bool is_number_type(const Type *t)
 {
     return is_integer_type(t) || t->kind == TYPE_DOUBLE;
 }
+
+bool is_float_type(const Type *t)
+{
+    return is_integer_type(t) || t->kind == TYPE_FLOAT;
```
Akuli, line 142: Integers shouldn't be considered floats. I see this function as something that checks whether you have `float` and `double`, similar to how `is_integer_type()` checks for `byte`/`int`/`long`.
```suggestion
    return t->kind == TYPE_FLOAT || t->kind == TYPE_DOUBLE;
```

Akuli (commented):
This PR is quite long. Maybe we should split it up into two smaller PRs:
- The `float` data type, should be usable with e.g. `x: float = 1` or `x = 12.34 as float`
- `12.34f` syntax to construct floats without requiring the `as` keyword

src/jou_compiler.h
littlewhitecloud: Merge double_text and float_text? 
Seems like a good idea.

src/jou_compiler.h
Akuli: Actually, I'm no longer sure about this. We will only ever have two float-like types anyway, but we have many different kinds of integers.

It's fine in both ways :)

littlewhitecloud: So, we just need TYPE_DOUBLE?

src/jou_compiler.h
littlewhitecloud: But in  file "typecheck.c" line 268 ~ 272, how to we identify float or double?
```c
    if (got_numbers && ???) { // What to check?
        cast_type = floatType;
    }
    if (got_numbers && !got_integers) {
        cast_type = doubleType;
    }
```

littlewhitecloud: if not got_integers is double, so how do we split up double/float?

src/jou_compiler.h
Akuli: That would be one way. But as I said, it's also fine to keep it like you have now and have `TYPE_FLOAT` and `TYPE_DOUBLE`.

src/jou_compiler.h
Akuli: As I mentioned in another comment, I think you should use `double` whenever the left side or the right side is a double.

```c
if (got_numbers && !got_integers) {
    cast_type = (lhstypes->type == doubleType || rhstypes->type == doubleType) ? doubleType : floatType;
}
```

littlewhitecloud:
Fine, I will rebuild them, with latest version(the long type was added)

src/jou_compiler.h
littlewhitecloud: ok

littlewhitecloud:
...But I rebuilded it in a new pull request

PR #179, opened 2023-02-01
Run tests in a directory that contains a space

### LINKED ISSUE #176: Running tests with space in folder name
GitHub actions runs its commands in a folder whose name doesn't contain spaces. If we move to a folder with a space in the name, we might discover more bugs similar  to #165.

Akuli:
Fixes #176

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -123,7 +123,7 @@ jobs:
       - run: mkdir "test dir"
       - run: mv jou/* repo/tests repo/runtests.sh "test dir"
         shell: bash
-      - run: "test dir\jou.exe" --verbose "test dir\examples\hello.jou"
+      - run: .\jou.exe --verbose "test dir/examples/hello.jou"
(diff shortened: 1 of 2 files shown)


.github/workflows/windows.yml
```diff
@@ -119,10 +119,13 @@ jobs:
...
       - run: unzip jou.zip
-      - run: mv jou/* repo/tests repo/runtests.sh .
+      # Add a space in the folder name to trigger bugs like #165
+      - run: mkdir "test dir"
+      - run: mv jou/* repo/tests repo/runtests.sh "test dir"
         shell: bash
-      - run: .\jou.exe --verbose examples/hello.jou
-      - run: ./runtests.sh
+      - run: cd "test dir" && ./jou.exe --verbose examples/hello.jou
+        shell: bash
```
Akuli, line 127: Let's just try running it inside the subdirectory. It shouldn't rely on the current directory.

```suggestion
      - run: .\jou.exe --verbose "test dir/examples/hello.jou"
```

PR #184, opened 2023-02-01
Global variables

### LINKED ISSUE #150: Global variables
Often a bad idea, sometimes a good idea. Needed for `fprintf()`.

To make global variables stand out, I think I will add a `global` keyword that can be used like so

```python
global foo: int

def set_foo(value: int) -> void:
    foo = value

def get_foo() -> int:
    return foo
```

It should also be possible to declare a global variable that is defined elsewhere. For example, `stdlib/io.jou` could do:
```python
declare global stdin: FILE*
declare global stdout: FILE*
declare global stderr: FILE*
```

Akuli:
Fixes #150

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -88,6 +88,35 @@ static AstType parse_type(const Token **tokens)
     return result;
 }
 
+// does not eat a trailing newline
+static AstVarDeclaration parse_var_declaration(const Token **tokens, const char *expected_what_for_name)
+{
+    AstVarDeclaration result;
+
+    if ((*tokens)->type != TOKEN_NAME) {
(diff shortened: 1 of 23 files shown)


src/parse.c
```diff
@@ -609,6 +609,27 @@ static AstIfStatement parse_if_statement(const Token **tokens)
     };
 }
 
+// does not eat a trailing newline
+static AstVarDeclaration parse_var_declaration(const Token **tokens)
+{
+    AstVarDeclaration result;
+
+    safe_strcpy(result.name, (*tokens)->data.name);
+    *tokens += 2;
```
Akuli, line 618: Should verify that there is `:`

PR #190, opened 2023-02-01
Refactor code to use more lists and AstNameTypeValue

Akuli:


src/parse.c
```diff
@@ -125,27 +125,6 @@ static_assert(sizeof(struct Name) == 100, "u have weird c compiler...");
 typedef List(struct Name) NameList;
 typedef List(AstType) TypeList;
```
Akuli, line 126: maybe these can be deleted too

PR #199, opened 2023-02-04
Fix explicit casts involving floats/doubles

### LINKED ISSUE #156: Casting from double to integer
```python3
from "stdlib/io.jou" import printf

def main() -> int:
    foo = 12.34
    printf("%d\n", foo as int)
    return 0
```

Expected output: 12

Actual output: `compiler error in file "a.jou", line 5: cannot cast from type double to int`

Presumably this should require an *explicit* conversion, so `foo: int = 12.34` should be an error.

Part of #151

### LINKED ISSUE #195: Explicit cast from double to float doesn't work
```
from "stdlib/io.jou" import printf

def main() -> int:
    x = 12.34 as float
    printf("%.2f\n", x)
    return 0
```

Expected output: `12.34`
Actual output: `compiler error in file "asd.jou", line 4: cannot cast from type double to float`

Akuli:
Fixes #156 
Fixes #195

### DIFF
--- a/src/codegen.c
+++ b/src/codegen.c
@@ -278,17 +278,19 @@ static void codegen_instruction(const struct State *st, const CfInstruction *ins
                         // same size, LLVM doesn't distinguish signed and unsigned integer types
                         setdest(getop(0));
                     }
-                } else if ((is_integer_type(from) && to->kind == TYPE_DOUBLE) || (is_integer_type(from) && to->kind == TYPE_FLOAT)) {
+                } else if (is_integer_type(from) && (to->kind == TYPE_DOUBLE || to->kind == TYPE_FLOAT)) {
                     // integer --> double / float
                     if (from->kind == TYPE_SIGNED_INTEGER)
                         setdest(LLVMBuildSIToFP(st->builder, getop(0), codegen_type(to), "cast"));
                     else
                         setdest(LLVMBuildUIToFP(st->builder, getop(0), codegen_type(to), "cast"));
+                } else if ((from == floatType || from == doubleType) && is_integer_type(to)) {
+                    if (to->kind == TYPE_SIGNED_INTEGER)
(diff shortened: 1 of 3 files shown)


src/codegen.c
```diff
@@ -278,17 +278,19 @@ static void codegen_instruction(const struct State *st, const CfInstruction *ins
...
+                } else if ((from == floatType || from == doubleType) && is_integer_type(to)) {
+                    if (to->kind == TYPE_SIGNED_INTEGER)
+                        setdest(LLVMBuildFPToSI(st->builder, getop(0), codegen_type(to), "cast"));
+                    else
+                        setdest(LLVMBuildFPToUI(st->builder, getop(0), codegen_type(to), "cast"));
                 } else if ((from == floatType || from == doubleType) && (to == floatType || to == doubleType)) {
                     setdest(LLVMBuildFPCast(st->builder, getop(0), codegen_type(to), "cast"));
-                } else {
-                    // TODO: double/float --> integer
-                    assert(0);
```
Akuli, line 291: Should not be deleted.

tests/should_succeed/as.jou
```diff
@@ -11,4 +11,22 @@ def main() -> int:
...
+    printf("%d\n", 12.34 as int)  # Output: 12
+    printf("%d\n", 12.99 as int)  # Output: 12
+    printf("%d\n", -12.34 as int)  # Output: -12
+    printf("%d\n", -12.99 as int)  # Output: -12
+
+    printf("%d\n", 12.34 as byte)  # Output: 12
+    printf("%d\n", 12.99 as int)  # Output: 12
+
+    # The following two lines are UB and produce different
+    # values depending on the -O3 flag:
```
Akuli, line 28:
```suggestion
    # values with -O3 and without:
```

PR #206, opened 2023-02-09
Makefile: Fix clean rule to delete tmp from the right place

Akuli:


.github/workflows/linux.yml
```diff
@@ -19,3 +19,9 @@ jobs:
     - run: ./runtests.sh --verbose './jou ${{ matrix.opt-level }} --verbose %s'
     - run: ./runtests.sh --verbose --valgrind './jou ${{ matrix.opt-level }} %s'
     # valgrind+verbose isn't meaningful: test script would ignore valgrind output
+    # Make sure that "make clean" cleans everything
+    - run: |
+        if [ "$(git status --porcelain)" != "" ]; then
+          git status
```
Akuli, line 25:
```suggestion
        if [ "$(git status --porcelain --ignored)" != "" ]; then
          git status --ignored
```

.github/workflows/linux.yml
```diff
@@ -19,3 +19,9 @@ jobs:
     - run: ./runtests.sh --verbose './jou ${{ matrix.opt-level }} --verbose %s'
     - run: ./runtests.sh --verbose --valgrind './jou ${{ matrix.opt-level }} %s'
     # valgrind+verbose isn't meaningful: test script would ignore valgrind output
+    # Make sure that "make clean" cleans everything
+    - run: |
```
Akuli, line 23:
```suggestion
    - name: 'Check that "make clean" cleans all files not committed to Git'
      run: |
```

PR #217, opened 2023-02-12
math lib

littlewhitecloud:
Add mathlib for jou
(sorry, my now repo doesn’t update to the old pull request, so l created a new one)


littlewhitecloud:
Successed

Akuli:
A better solution would be to make a new branch before you edit the files. The problem is that you make all your pull requests from your `main` branch.

stdlib/math.jou
```diff
@@ -0,0 +1,41 @@
+declare abs(x: int) -> int
+declare llabs(x: long) -> long # Note: it works on the latest verion of jou
```
Akuli, line 2:
```suggestion
declare llabs(x: long) -> long
```

stdlib/math.jou
```diff
@@ -0,0 +1,41 @@
...
+declare fabs(x: double) -> double
+
+declare ceil(x: double) -> double
+declare floor(x: double) -> double
+declare round(x: double) -> double
+
+declare acos(x: double) -> double
+declare asin(x: double) -> double
+declare atan(x: double) -> double
+declare atan2(x: double, y: double) -> double
```
Akuli, line 12: The `atan2` function returns the angle of a point, given its x and y coordinates. It expects the y coordinate first, so let's name the arguments accordingly.
```suggestion
declare atan2(y: double, x: double) -> double
```

stdlib/math.jou
```diff
@@ -0,0 +1,41 @@
...
+declare cbrt(x: double) -> double
+
+declare log(x: double) -> double
+declare log10(x: double) -> double
+declare log2(x: double) -> double
+
+declare exp(x: double) -> double
+declare exp2(x: double) -> double
+declare pow(x: double, y: double) -> double
+declare pow10(p: int) -> double
```
Akuli, line 28: We should probably use `exp10` instead of `pow10`. The manual page of `pow10` says:

```
       Note  well:  These functions perform exactly the same task as the func‐
       tions described in exp10(3), with the difference that the latter  func‐
       tions  are now standardized in TS 18661-4:2015.  Those latter functions
       should be used in preference to the functions described in this page.
```

Also, please add a test for all functions. There is no test for `pow10`.

Akuli:
Turns out that neither `pow10` nor `exp10` work on Windows. But that's fine: we don't really need them since you can do `pow(10, x)` anyway.

Akuli:
Thanks for working on this!

PR #220, opened 2023-02-13
Add warnings for unused imports

### LINKED ISSUE #218: unused import warning
```
from "stdlib/io.jou" import printf, puts

def main() -> int:
    puts("hey")
    return 0
```

Expected result: warning on line 1 because `printf` is not used.
Actual result: no warning

Akuli:
Fixes #218

### DIFF
--- a/src/codegen.c
+++ b/src/codegen.c
@@ -438,12 +438,12 @@ LLVMModuleRef codegen(const CfGraphFile *cfgfile, const TypeContext *typectx)
     for (GlobalVariable **v = typectx->globals.ptr; v < End(typectx->globals); v++) {
         LLVMTypeRef t = codegen_type((*v)->type);
         LLVMValueRef globalptr = LLVMAddGlobal(st.module, t, (*v)->name);
-        if (!(*v)->defined_outside_jou)
+        if (!(*v)->defined_outside_current_file)
             LLVMSetInitializer(globalptr, LLVMGetUndef(t));
     }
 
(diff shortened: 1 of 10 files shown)


src/codegen.c
```diff
@@ -438,12 +438,12 @@ LLVMModuleRef codegen(const CfGraphFile *cfgfile, const TypeContext *typectx)
     for (GlobalVariable **v = typectx->globals.ptr; v < End(typectx->globals); v++) {
         LLVMTypeRef t = codegen_type((*v)->type);
         LLVMValueRef globalptr = LLVMAddGlobal(st.module, t, (*v)->name);
-        if (!(*v)->defined_outside_jou)
+        if (!(*v)->defined_outside_current_file)
```
Akuli, line 441: Should maybe change this to its opposite and name it `defined_in_current_file`

src/jou_compiler.h
```diff
@@ -408,7 +408,8 @@ Signature copy_signature(const Signature *sig);
 struct GlobalVariable {
     char name[100];  // Same as in user's code, never empty
     const Type *type;
-    bool defined_outside_jou;  // true for variables like stdout
+    bool defined_outside_current_file;  // imported variables, also external variables like stdout
+    bool *usedptr;
```
Akuli, line 412:
```suggestion
    bool *usedptr;  // If non-NULL, set to true when the variable is used. This is how we detect unused imports.
```

PR #221, opened 2023-02-13
Add a --linker-flags argument

### LINKED ISSUE #214: External dependencies in Jou programs
This may be the last thing I need to do before I can start working on a self-hosted compiler (#116).

Let's say you want to write a C or Jou program that uses LLVM:
- On Windows, you give a path to `LLVM-C.lib` as a command-line argument to the linker, and you make sure that a somewhat large collection of DLL files is available when the program runs. You need to download these files by either running LLVM's installer or by extracting them from LLVM's installer without running the installer.
- On Linux, before you compile the program, you need to run `sudo apt install llvm-11-dev` on terminal to install LLVM. Then you can run a terminal command `llvm-config-11 --ldflags --libs` (where `11` is the LLVM version), which prints something like `-L/usr/lib/llvm-11/lib -lLLVM-11`. These are flags that you need to give to the linker.

In other words: it is a mess, and what exactly you need to do depends on your OS. Maybe some day there will be a Jou package manager that handles the mess for you.

But for now, I think it would be good enough if you could specify linker flags in Jou code. You still need to download/install the external library some os-specific way, but at least it would be possible to use it. Maybe a `linker_flags` keyword? Something like this:

```python3
if WINDOWS:  # Code inside this would be compiled only on windows.
    linker_flags "lib/LLVM-C.lib"
else:
    linker_flags "-L/usr/lib/llvm-11/lib -lLLVM-11"

declare LLVMSomeFunction(...)

def main() -> int:
    LLVMSomeFunction(...)
    return 0
```

Akuli:
A bit annoying to use, but fixes #214 for now. Hopefully I will eventually create an actual package manager.

### DIFF
--- a/examples/x11_window.jou
+++ b/examples/x11_window.jou
@@ -1,0 +1,58 @@
+# To run this program, you need linux:
+#
+#   $ ./jou --linker-flags "-lX11" examples/x11_window.jou
+#
+
+declare usleep(x: int) -> int
+
+struct Display:
+    _dummy: int
+struct XGCValues:
+    _dummy: int
+struct GC:
+    _dummy: int
+
+declare XOpenDisplay(name: byte*) -> Display*
+declare XCreateSimpleWindow(
+    display: Display*,
+    parent: long,
+    x: int,
+    y: int,
+    width: int,
+    height: int,
+    border_width: int,
+    border: long,
+    background: long,
+) -> long
+declare XCreateGC(display: Display*, drawable: long, valuemask: long, values: XGCValues*) -> GC*
+declare XSetForeground(display: Display*, gc: GC*, foreground: long) -> int
+declare XSelectInput(display: Display*, window: long, event_mask: long) -> int
+declare XMapRaised(display: Display*, window: long) -> int
+declare XDrawImageString(
+    display: Display*,
+    drawable: long,
+    gc: GC*,
+    x: int,
+    y: int,
+    string: byte*,
+    length: int,
+) -> int
+declare XDefaultRootWindow(display: Display*) -> long
+declare XStoreName(display: Display*, window: long, name: byte*) -> int
+declare XFlush(display: Display*) -> int
+
+def main() -> int:
+    display = XOpenDisplay("")
+    window = XCreateSimpleWindow(display, XDefaultRootWindow(display), 200, 200, 200, 200, 5, 0, 0xff00ffL)
+    XStoreName(display, window, "Hello pink world")
+
+    gc = XCreateGC(display, window, 0, NULL)
+    XSetForeground(display, gc, 0xffffffL)
+
+    XSelectInput(display, window, 32768)
+    XMapRaised(display, window)
+
+    while True:
+        usleep(100)
+        x = XDrawImageString(display, window, gc, 50, 50, "hello", 5)
+        XFlush(display)  # This makes the program die when closed. Don't know why :)
--- a/src/main.c
+++ b/src/main.c
@@ -34,7 +34,7 @@ static void optimize(LLVMModuleRef module, int level)
 
 static const char help_fmt[] =
     "Usage:\n"
-    "  <argv0> [-o OUTFILE] [-O0|-O1|-O2|-O3] [--verbose] FILENAME\n"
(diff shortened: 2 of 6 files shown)


src/main.c
```diff
@@ -75,6 +76,19 @@ static void parse_arguments(int argc, char **argv, CommandLineFlags *flags, cons
         } else if (!strcmp(argv[i], "--verbose")) {
             flags->verbose = true;
             i++;
+        } else if (!strcmp(argv[i], "--linker-flags")) {
+            if (flags->linker_flags) {
+                // TODO: test
```
Akuli, line 81:
```suggestion
```

src/main.c
```diff
@@ -75,6 +76,19 @@ static void parse_arguments(int argc, char **argv, CommandLineFlags *flags, cons
...
+                fprintf(stderr, "%s: --linker-flags cannot be given multiple times", argv[0]);
+                goto wrong_usage;
+            }
+            if (argc-i < 2) {
+                // TODO: test
```
Akuli, line 86:
```suggestion
```

examples/x11_window.jou
```diff
@@ -0,0 +1,58 @@
...
+
+    gc = XCreateGC(display, window, 0, NULL)
+    XSetForeground(display, gc, 0xffffffL)
+
+    XSelectInput(display, window, 32768)
+    XMapRaised(display, window)
+
+    while True:
+        usleep(100)
+        x = XDrawImageString(display, window, gc, 50, 50, "hello", 5)
```
taahol, line 57: Variable `x` not needed.

Akuli: Related: #26

tests/should_succeed/compiler_cli.jou
```diff
@@ -28,6 +28,8 @@ def main() -> int:
     run_jou("-O8 examples/hello.jou")  # Output: <jouexe>: unknown argument "-O8" (try "<jouexe> --help")
     run_jou("--lolwat")  # Output: <jouexe>: unknown argument "--lolwat" (try "<jouexe> --help")
     run_jou("lolwat.jou")  # Output: compiler error in file "lolwat.jou": cannot open file: No such file or directory
+    run_jou("--linker-flags")  # Output: <jouexe>: there must be a string of flags after --linker-flags (try "<jouexe> --help")
+    run_jou("--linker-flags x --linker-flags y")  # Output: <jouexe>: --linker-flags cannot be given multiple times (try "./jou --help")
```
littlewhitecloud, line 32: # Output: <jouexe>: --linker-flags cannot be given multiple times (try "<jouexe> --help")

tests/should_succeed/compiler_cli.jou
```diff
@@ -27,16 +28,19 @@ def main() -> int:
...
```
littlewhitecloud, line 32: run_jou("--linker-flags x --linker-flags y")  # Output: <jouexe>: --linker-flags cannot be given multiple times (try "./jou --help")

littlewhitecloud, line 32: Should be:
```python
run_jou("--linker-flags x --linker-flags y")  # Output: <jouexe>: --linker-flags cannot be given multiple times (try "<jouexe> --help")
```

littlewhitecloud (changes requested):

tests/should_succeed/compiler_cli.jou
littlewhitecloud, line 32: Should be:
```python
run_jou("--linker-flags x --linker-flags y")  # Output: <jouexe>: --linker-flags cannot be given multiple times (try "<jouexe> --help")
```

littlewhitecloud, line 32:
```suggestion
    run_jou("--linker-flags x --linker-flags y")  # Output: <jouexe>: --linker-flags cannot be given multiple times (try "<jouexe> --help")
```

littlewhitecloud (changes requested):

tests/should_succeed/compiler_cli.jou
Akuli: Thanks :)

PR #222, opened 2023-02-17
From one file import all functions

littlewhitecloud:
Just support syntax like this:
```python
from ... import *
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -323,7 +323,7 @@ static void add_imported_symbols(struct CompileState *compst)
             assert(from);
 
             for (struct ExportSymbol *es = from->pending_exports; es->name[0]; es++) {
-                if (es->name == '*') {
+                if (!strcmp(es->name, "*")) {
                     printf("WOW, it successed!                                             sd");
                 }
                 if (!strcmp(imp->symbolname, es->name)) {
--- a/src/main.c
+++ b/src/main.c
@@ -323,6 +323,9 @@ static void add_imported_symbols(struct CompileState *compst)
             assert(from);
 
             for (struct ExportSymbol *es = from->pending_exports; es->name[0]; es++) {
+                if (es->name == '*') {
+                    printf("WOW, it successed!                                             sd");
+                }
                 if (!strcmp(imp->symbolname, es->name)) {
                     if (compst->flags.verbose) {
                         const char *kindstr;
--- a/src/parse.c
+++ b/src/parse.c
@@ -854,9 +854,14 @@ static void parse_import(const Token **tokens, const char *stdlib_path, Toplevel
 
     bool parens = is_operator(*tokens, "(");
     if(parens) ++*tokens;
+    /*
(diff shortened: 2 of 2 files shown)


littlewhitecloud:
just draft now, there is more things I need to do
- [x] Support syntax
- [ ] No error 
- [ ] Support import
- [ ] Check

Akuli:
I'm not sure if I want this syntax. In languages that have it, it is usually considered bad style.

src/parse.c
```diff
@@ -853,10 +853,15 @@ static void parse_import(const Token **tokens, const char *stdlib_path, Toplevel
...
+    if(parens) ++*tokens;
+    /*
+    if (!strcmp((*tokens)->data.name, '*') && (*tokens)->type != TOKEN_NAME)
+        printf("%d", !strcmp((*tokens)->data.name, '*'));
+        printf("\n");
+    */
 
     do {
-        if ((*tokens)->type != TOKEN_NAME)
+        if ((*tokens)->type != TOKEN_NAME && (*tokens)->type != TOKEN_OPERATOR) // Hah, failed many times make it to check *(tokens)->data.name == '*' successed
```
Akuli, line 864: You can use the `is_operator()` function:
```suggestion
        if ((*tokens)->type != TOKEN_NAME && !is_operator(*tokens, "*"))
```

src/main.c
```diff
@@ -320,9 +320,12 @@ static void add_imported_symbols(struct CompileState *compst)
         for (AstToplevelNode *ast = to->ast; ast->kind == AST_TOPLEVEL_IMPORT; ast++) {
             AstImport *imp = &ast->data.import;
             struct FileState *from = find_file(compst, imp->path);
-            assert(from);
+            assert(from);
 
-            for (struct ExportSymbol *es = from->pending_exports; es->name[0]; es++) {
+            for (struct ExportSymbol *es = from->pending_exports; es->name[0]; es++) {
+                if (es->name == '*') {
```
Akuli, line 326: Maybe:

```suggestion
                if (!strcmp(es->name, "*")) {
```

src/main.c
```diff
@@ -320,9 +320,12 @@ static void add_imported_symbols(struct CompileState *compst)
...
             struct FileState *from = find_file(compst, imp->path);
-            assert(from);
+            assert(from);
 
-            for (struct ExportSymbol *es = from->pending_exports; es->name[0]; es++) {
+            for (struct ExportSymbol *es = from->pending_exports; es->name[0]; es++) {
+                if (es->name == '*') {
+                    printf("WOW, it successed!                                             sd");
+                }
                 if (!strcmp(imp->symbolname, es->name)) {
```
Akuli, line 329: I think this is the line that should be changed. If we are importing a symbol named `*`, it means we should match all exported symbols.

```suggestion
                if (!strcmp(imp->symbolname, es->name) || !strcmp(imp->symbolname, "*")) {
```

Akuli:
Thanks for working on this. I might merge it later if I decide it is a good idea.

PR #225, opened 2023-02-18
Use more typecheck_expression_not_void

### LINKED ISSUE #223: Using void return value segfaults compiler
```
def foo() -> void:
    return

def bar() -> void:
    x = foo()
```

Expected result: user-friendly error message

Akuli:
Fixes #223 

First commit created on my phone lol


Akuli:
I did the entire pr on my phone.

PR #240, opened 2023-02-25
Set default optimize to 1

littlewhitecloud:
Set default optimize level to 1
Default optimize level with no optimize level set:
image
Set optimize level to 3 and it will overwrite the default level:
image


Akuli:
I'm not convinced this is a good idea. For example, consider this program (taken from doc/perf.md):

```
from "stdlib/io.jou" import printf, getchar

def main() -> int:
    printf("Crash this program? (y/n) ")
    if getchar() == 'y':
        foo: int* = NULL
        x = *foo
    return 0
```

It crashes as expected with `-O0`, but doesn't crash with `-O1` because the compiler assumes you don't access a null pointer. I think this isn't the behaviour you should get, unless you specifically ask Jou to optimize the code.

As I mentioned in the issue, the generated LLVM IR is quite dumb (until optimized). It contains a lot of `alloca` instructions, even for things that clearly don't need an `alloca`.

Akuli:
After thinking about this a bit more, I think it is a good solution for now. I will make the self-hosted compiler so that it generates better LLVM IR, and for the compiler we have now, this is useful because it means bigger programs compile quickly.

But before merging, we should update `doc/perf.md`.

littlewhitecloud:
Maybe we can add a keyword to tell the compiler don't optmize here

littlewhitecloud:
Maybe...

littlewhitecloud:
> After thinking about this a bit more, I think it is a good solution for now. I will make the self-hosted compiler so that it generates better LLVM IR, and for the compiler we have now, this is useful because it means bigger programs compile quickly.
> 
> But before merging, we should update .`doc/perf.md`

What should I changed in `.doc/perf.md`?

littlewhitecloud:
changed O1 to default in the table and add a ask & answer in perf.md

littlewhitecloud:
What should L do now?

Akuli:
I edited the documentation a bit more: there were still other parts of it that claimed `-O0` is the default. I think this is ready to be merged once the checks are green.

That said, I plan to write the self-hosted compiler so that it works well with `-O0`, and then make `-O0` the default again. This PR is still a good temporary solution.

Akuli:
Sorry about taking so long to get this merged! I was too excited about the self-hosted compiler :)

PR #241, opened 2023-02-25
Methods

### LINKED ISSUE #44: Should structs have methods?
In this context, "class" means "struct with methods".

C doesn't have methods in structs, but a lot of people would like to have a "C with classes" language with most of the minimality of C. This was historically C++ but definitely isn't anymore.

Another advantage of classes is the similarity to Python: all Python developers know classes and expect to see them in a language, and the `class` keyword syntax highlights nicely when abusing Python's syntax highlighting.

A disadvantage with classes is that they easily lead to a lot of unnecessary fluff that doesn't belong in a minimal language (like C or Jou):
- Generics
- Templates
- Inheritance and other fancy/disgusting OOP tricks
- Runtime type info
- Static methods / class methods / extrafancy abstract boilerplate poopoo methods
- ...

A bigger disadvantage of classes is that when you have classes, it becomes tempting to make wrappers everything standard library. If you want to stay close to C (like C++ did and Jou does), you end up with two copies of everything. For example, in C++ there is `std::cin` and `stdin`, `std::string` and `char*`, smart pointers and C pointers, etc.

One more disadvantage is magic methods that can hide a lot, such as copy constructors, destructors and operator overloading. Ewww.

I think I will add classes to Jou, but intentionally restrict them a lot to prevent these problems:
- **No new standard-library classes.** For example, there will be no `std::string` that hides all the fun details inside itself. Jou should be a language where you enjoy playing with pointers, not a language where you abstract everything away until your code is unreadable and then blame other people for not understanding it. (Looking at you, C++)
- **Wrapping C code is painless, and doesn't involve making new structs/classes.** Below is an idea about how that could work.

```
# io.jou
opaque struct FILE:  # opaque = fields not known to compiler
    functionmethod fprintf(self: FILE*, format: byte*, ...) -> int
    functionmethod fclose(self: FILE*)



# asdasd.jou
import io

def main() -> int:
    FILE *file = fopen("hello.txt", "w")
    file->fprintf("hello\n")
    file->fclose()
    return 0
```

Akuli:
Fixes #44


src/build_cfg.c
```diff
@@ -415,25 +415,27 @@ static const LocalVariable *build_address_of_expression(struct State *st, const
     assert(0);
 }
 
-static const LocalVariable *build_function_call(struct State *st, const AstExpression *expr)
+static const LocalVariable *build_function_call(struct State *st, const Location location, const AstCall *call, const LocalVariable *self, const Type *returntype)
```
Akuli, line 418: Should renamme build_function_or_method_call

src/build_cfg.c
```diff
@@ -415,25 +415,27 @@ static const LocalVariable *build_address_of_expression(struct State *st, const
...
 {
-    assert(expr->kind == AST_EXPR_FUNCTION_CALL);
+    const LocalVariable **args = calloc(call->nargs + 2, sizeof(args[0]));  // NOLINT
+    int k = 0;
+    if (self)
+        args[k++] = self;
```
Akuli, line 423: Ideally the self argument could go anywhere, e.g. `fread()` takes it as the last argument.

src/build_cfg.c
```diff
@@ -471,8 +473,17 @@ static const LocalVariable *build_expression(struct State *st, const AstExpressi
     const LocalVariable *result, *temp;
 
     switch(expr->kind) {
+    case AST_EXPR_DEREF_AND_CALL_METHOD:
+        assert(0);
```
Akuli, line 477:
```suggestion
        assert(0);  // TODO: implement
```

src/typecheck.c
```diff
@@ -631,17 +632,40 @@ static const char *nth(int n)
...
+
     char *sigstr = signature_to_string(sig, false);
 
-    if (call->nargs < sig->nargs || (call->nargs > sig->nargs && !sig->takes_varargs)) {
+    int n = call->nargs + !!self_type;
+    if (n < sig->nargs || (n > sig->nargs && !sig->takes_varargs)) {
         fail_with_error(
             location,
-            "function %s takes %d argument%s, but it was called with %d argument%s",
+            "%s %s takes %d argument%s, but it was called with %d argument%s",
```
Akuli, line 667: Should add a test for this on methods. And probably for everything else that mentions a "function", as that should automagically say "method" instead

src/typecheck.c
```diff
@@ -802,6 +829,12 @@ static ExpressionTypes *typecheck_expression(TypeContext *ctx, const AstExpressi
                 temptype->name);
         result = typecheck_struct_field(temptype->data.valuetype, expr->data.structfield.fieldname, expr->location);
         break;
+    case AST_EXPR_CALL_METHOD:
+        temptype = typecheck_expression_not_void(ctx, expr->data.methodcall.obj)->type;
+        result = typecheck_function_or_method_call(ctx, &expr->data.methodcall.call, temptype, expr->location);
+        break;
+    case AST_EXPR_DEREF_AND_CALL_METHOD:
+        assert(0);
```
Akuli, line 837:
```suggestion
        assert(0);  // TODO
```

src/typecheck.c
```diff
@@ -1064,6 +1097,24 @@ const Signature *typecheck_function_body(TypeContext *ctx, const char *name, con
...
+    assert(structtype);
+
+    for (AstFunctionDef *m = structdef->methods.ptr; m < End(structdef->methods); m++) {
+        char name[sizeof structtype->name + sizeof "." + sizeof m->signature.funcname];
+        sprintf(name, "%s.%s", structtype->name, m->signature.funcname);
+        typecheck_function_body(ctx, name, &m->body);
+        reset_type_context(ctx);
+    }
+}
+
```
Akuli, line 1117:
```suggestion
```

PR #251, opened 2023-02-25
Use methods in the self-hosted compiler

Akuli:
This code was written before Jou had methods.

### DIFF
--- a/self_hosted/errors_and_warnings.jou
+++ b/self_hosted/errors_and_warnings.jou
@@ -5,18 +5,18 @@ struct Location:
     path: byte*  # Not owned. Points to a string that is held elsewhere.
     lineno: int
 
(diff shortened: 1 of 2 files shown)


self_hosted/errors_and_warnings.jou
```diff
@@ -5,18 +5,18 @@ struct Location:
     path: byte*  # Not owned. Points to a string that is held elsewhere.
     lineno: int
 
-def fail(location: Location, message: byte*) -> void:
-    # When stdout is redirected to same place as stderr,
-    # make sure that normal printf()s show up before our error.
-    fflush(stdout)
-    fflush(stderr)
+    def fail(self: Location, message: byte*) -> void:
```
Akuli, line 8: I don't like this as a method

PR #253, opened 2023-02-25
add nicer array syntax

### LINKED ISSUE #144: array syntax
https://github.com/Akuli/jou/blob/3ba82086d4a03c2d0c198a6afa01286c44094183/tests/should_succeed/array.jou#L6-L10

Would be better if I could do just `return [4, 5, 6]`

Akuli:
Fixes #144

### DIFF
--- a/src/typecheck.c
+++ b/src/typecheck.c
@@ -310,15 +310,10 @@ static noreturn void fail_with_implicit_cast_error(
     fail_with_error(location, "%.*s", msg.len, msg.ptr);
 }
 
-static void do_implicit_cast(
-    ExpressionTypes *types, const Type *to, Location location, const char *errormsg_template)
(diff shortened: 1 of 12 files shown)


src/typecheck.c
```diff
@@ -815,6 +820,62 @@ static bool enum_member_exists(const Type *t, const char *name)
...
+                break;
+            }
+        }
+        if (t_compatible_with_all_others) {
+            /*
+            This assertion should never fail, assuming that
+
+                can_cast_implicitly(A,B) && can_cast_implicitly(B,A)
+
+            is possible only when A and B are the same type.
```
Akuli, line 855: This assumption isn't always true, consider `int*` and `void*`. A list containing both of these should probably have item type `void*`, since it would turn into a `void*` list when a new item is added.

PR #256, opened 2023-02-25
Tokenize escapes according to doc/syntax-spec.md

Akuli:
I fixed/updated both tokenizers.


self_hosted/tokenizer.jou
```diff
@@ -209,6 +209,56 @@ struct Tokenizer:
...
+                c = '\\'
+            elif after_backslash == '\'':
+                c = '\''
+            elif after_backslash == '"':
+                fail(self->location, "double quotes shouldn't be escaped in byte literals")
+            elif '0' <= after_backslash and after_backslash <= '9':
+                c = after_backslash - '0'
+            elif after_backslash < 0x80 and isprint(after_backslash) != 0:
+                message: byte* = malloc(100)
+                sprintf(message, "unknown escape: '\\%c'", after_backslash)
```
Akuli, line 242: I wonder if this runs when there is a backslash at the end of a line.

Akuli: I fixed it, but it's difficult to test automatically: If I add a `# Error` comment at the end of the line, then the backslash is no longer at the end of a line.

PR #260, opened 2023-02-26
github actions: Avoid running valgrind when it is unnecessary

Akuli:
Valgrinding is slow.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -18,7 +18,18 @@ jobs:
       - run: LLVM_CONFIG=llvm-config-${{ matrix.llvm-version }} make
       - run: ./runtests.sh --verbose './jou ${{ matrix.opt-level }} %s'
       - run: ./runtests.sh --verbose './jou ${{ matrix.opt-level }} --verbose %s'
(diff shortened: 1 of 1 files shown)


.github/workflows/linux.yml
```diff
@@ -18,7 +18,18 @@ jobs:
       - run: LLVM_CONFIG=llvm-config-${{ matrix.llvm-version }} make
       - run: ./runtests.sh --verbose './jou ${{ matrix.opt-level }} %s'
       - run: ./runtests.sh --verbose './jou ${{ matrix.opt-level }} --verbose %s'
-      - run: ./runtests.sh --verbose --valgrind './jou ${{ matrix.opt-level }} %s'
+      # Valgrinding is slow. Do it only when this file or something in the compiler has been modified.
+      - name: Figure out if we need to run valgrind
+        id: check-need-valgrind
+        run: |
+          git fetch https://github.com/Akuli/jou main
+          if git diff --name-only FETCH_HEAD HEAD | grep -E '(linux.yml|\.c|\.h)$'; then
```
Akuli, line 26:
```suggestion
          if git diff --name-only FETCH_HEAD HEAD | grep -E '(linux\.yml|\.c|\.h)$'; then
```

PR #261, opened 2023-02-26
Add GitHub actions config to fix line endings automatically

### LINKED ISSUE #259: CRLF auto fix
littlewhitecloud often makes pull requests that contain files with CRLF line endings. I then run `dos2unix` on my computer to convert them to LF line endings, because that's usually what is committed to Git.

Ideally CI would convert the line endings automagically (except `tests/should_succeed/crlf.jou` which is intentionally committed with CRLF line endings).

Akuli:
Fixes #259


.github/workflows/linux.yml
```diff
@@ -5,6 +5,16 @@ on:
   pull_request:
 
 jobs:
+  autofix:
```
Akuli, line 8:
```suggestion
  fix-line-endings:
```

PR #266, opened 2023-02-26
codegen: Use i8* for all pointers inside structs

### LINKED ISSUE #264: self-referencing classes don't work
```python
class Foo:
    x: Foo*
    def foo(self: Foo*) -> void:
        return

def main() -> int:
    return 0
```

Expected result: no error

Akuli:
Fixes #264


src/codegen.c
```diff
@@ -244,7 +250,14 @@ static void codegen_instruction(const struct State *st, const CfInstruction *ins
                 int i = 0;
                 while (strcmp(classtype->data.classfields.names[i], ins->data.fieldname))
                     i++;
-                setdest(LLVMBuildStructGEP2(st->builder, codegen_type(classtype), getop(0), i, ins->data.fieldname));
+
+                LLVMValueRef val = LLVMBuildStructGEP2(st->builder, codegen_type(classtype), getop(0), i, ins->data.fieldname);
+                const Type *fieldtype = classtype->data.classfields.types[i];
+                if (fieldtype->kind == TYPE_POINTER) {
+                    // We lied to LLVM that the struct member is i8*, so that we can do self-referencing types
+                    val = LLVMBuildBitCast(st->builder, val, LLVMPointerType(codegen_type(fieldtype),0), "asd");
```
Akuli, line 258:
```suggestion
                    val = LLVMBuildBitCast(st->builder, val, LLVMPointerType(codegen_type(fieldtype),0), "struct_member_i8_hack");
```

PR #271, opened 2023-02-27
self-hosted compiler: Implement more parsing so that stdlib/io.jou parses

Akuli:
Will be needed for hello world

### DIFF
--- a/src/print.c
+++ b/src/print.c
@@ -211,8 +211,10 @@ static void print_ast_expression(const AstExpression *expr, struct TreePrinter t
         __attribute__((fallthrough));
     case AST_EXPR_CALL_METHOD:
         printf("call method \"%s\"\n", expr->data.methodcall.call.calledname);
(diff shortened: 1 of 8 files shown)


src/print.c
```diff
@@ -211,8 +216,10 @@ static void print_ast_expression(const AstExpression *expr, struct TreePrinter t
         __attribute__((fallthrough));
     case AST_EXPR_CALL_METHOD:
         printf("call method \"%s\"\n", expr->data.methodcall.call.calledname);
-        print_ast_expression(expr->data.methodcall.obj, print_tree_prefix(tp, false));
-        print_ast_call(&expr->data.methodcall.call, print_tree_prefix(tp, true));
+        struct TreePrinter self_printer = print_tree_prefix(tp, expr->data.methodcall.call.nargs == 0);
```
Akuli, line 219: these should be split into separate PR

PR #273, opened 2023-02-27
Fix various AST printing bugs

Akuli:
Extracted from #271. These aren't tested because it would be quite tedious.

### DIFF
--- a/src/print.c
+++ b/src/print.c
@@ -173,6 +173,11 @@ static void print_ast_function_signature(const AstSignature *sig)
         print_ast_type(&ntv->type);
         assert(!ntv->value);
     }
+    if (sig->takes_varargs) {
+        if (sig->args.len != 0)
+            printf(", ");
+        printf("...");
+    }
     printf(") -> ");
     print_ast_type(&sig->returntype);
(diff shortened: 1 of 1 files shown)


src/print.c
```diff
@@ -211,8 +216,10 @@ static void print_ast_expression(const AstExpression *expr, struct TreePrinter t
         __attribute__((fallthrough));
     case AST_EXPR_CALL_METHOD:
         printf("call method \"%s\"\n", expr->data.methodcall.call.calledname);
-        print_ast_expression(expr->data.methodcall.obj, print_tree_prefix(tp, false));
-        print_ast_call(&expr->data.methodcall.call, print_tree_prefix(tp, true));
+        struct TreePrinter self_printer = print_tree_prefix(tp, expr->data.methodcall.call.nargs == 0);
+        printf("self: ");
+        print_ast_expression(expr->data.methodcall.obj, self_printer);
+        print_ast_call(&expr->data.methodcall.call, tp);
```
Akuli, line 222: Maybe would be more clear for `print_ast_call()` to take the `self` expression.

PR #280, opened 2023-02-27
Improve self-hosted tokenizer

Akuli:
- Parse ints and longs in all supported bases (binary, octal, decimal, hex)
- Support space-ignoring mode (see syntax docs)
- Emit good error messages for mismatched parentheses

### DIFF
--- a/self_hosted/tokenizer.jou
+++ b/self_hosted/tokenizer.jou
@@ -101,6 +101,22 @@ def parse_integer(string: byte*, location: Location, nbits: int) -> long:
 
     return result
 
+def flip_paren(c: byte) -> byte:
+    if c == '(':
+        return ')'
+    if c == ')':
+        return '('
+    if c == '[':
+        return ']'
+    if c == ']':
+        return '['
+    if c == '{':
+        return '}'
+    if c == '}':
+        return '{'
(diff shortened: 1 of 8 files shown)


self_hosted/tokenizer.jou
```diff
@@ -38,6 +37,86 @@ def is_keyword(word: byte*) -> bool:
...
+    for i = 0; i < strlen(digits); i++:
+        if 'A' <= digits[i] and digits[i] <= 'F':
+            digit_value = 10 + (digits[i] - 'A')
+        elif 'a' <= digits[i] and digits[i] <= 'f':
+            digit_value = 10 + (digits[i] - 'a')
+        else:
+            digit_value = digits[i] - '0'
+
+        # Overflow isn't UB in Jou
+        if result * base < 0:
```
Akuli, line 84: This may not work correctly in bases other than 2, because an overflowing multiplication doesn't necessarily set the sign bit. But this should work:

```suggestion
        if (result * base) / base != result:
```

With any `base`, an overflowing multiplication will produce a very different value from the correct result, off by much more than `base-1`. The division cannot overflow, and it is affected by changes bigger than `base-1`.

PR #283, opened 2023-03-01
Type-checking for self-hosted compiler

Akuli:
Contains a couple temporary hacks, but this will be enough to run hello world :)


self_hosted/ast.jou
```diff
@@ -133,10 +153,10 @@ class AstNameTypeValue:
 
 class AstSignature:
     funcname_location: Location
-    funcname: byte[100]
+    funcname: byte[100]  # TODO: rename to function_or_method_name
```
Akuli, line 156: Might as well do now

self_hosted/main.jou
```diff
@@ -1,28 +1,208 @@
...
+
+enum CompilerMode:
+    TokenizeOnly    # Tokenize one file, don't recurse to imports
+    ParseOnly       # Tokenize and parse one file, don't recurse to imports
+    Compile         # Compile and link
+    CompileAndRun   # Compile, link and run a program (default)
+
+class CommandLineArgs:
+    mode: CompilerMode
+    output_file: byte*  # The argument after -o, possibly with .exe appended to it
```
Akuli, line 25: Should create a separate issue about adding `.exe`

self_hosted/runs_wrong.txt
```diff
@@ -0,0 +1,165 @@
...
+tests/syntax_error/missing_field_names.jou
+tests/syntax_error/missing_import_keyword.jou
+tests/syntax_error/missing_number_after_eminus.jou
+tests/syntax_error/missing_second_equal_sign.jou
+tests/syntax_error/multidot_float.jou
+tests/syntax_error/python_style_for.jou
+tests/syntax_error/self_outside_class.jou
+tests/syntax_error/triple_equals.jou
+tests/too_long/name.jou
+tests/wrong_type/arg.jou
```
Akuli, line 128: Why is `arg.jou` in the list?

self_hosted/runs_wrong.txt
```diff
@@ -0,0 +1,165 @@
...
+tests/wrong_type/inplace_add_doesnt_go_back.jou
+tests/wrong_type/int_to_enum.jou
+tests/wrong_type/mod.jou
+tests/wrong_type/neg.jou
+tests/wrong_type/not.jou
+tests/wrong_type/or.jou
+tests/wrong_type/plusplus.jou
+tests/wrong_type/pointer_assignment.jou
+tests/wrong_type/pointer_eq.jou
+tests/wrong_type/return_value.jou
```
Akuli, line 162: And why is  `return_value.jou` in this list?

self_hosted/typecheck.jou
```diff
@@ -0,0 +1,380 @@
...
+        ctx->current_function_signature = sig
+        typecheck_body(ctx, &ts->funcdef.body)
+        ctx->current_function_signature = NULL
+
+
+#
+#def main() -> int:
+#    init_types()
+#    printf("Hi\n")
+#    return 0
```
Akuli, line 380:
```suggestion
```

self_hosted/types.jou
```diff
@@ -0,0 +1,128 @@
...
+    strcpy(&long_type->name[0], "long")
+
+#def main() -> int:
+#    init_types()
+#    printf("Hi\n")
+#    printf("%s\n", &signed_integers[32].name[0])
+#    printf("%s\n", &unsigned_integers[32].name[0])
+#    printf("%s\n", &bool_type.name[0])
+#    printf("%s\n", &bool_type.get_pointer_type()->get_pointer_type()->name)
+#    return 0
```
Akuli, line 82:
```suggestion
```

self_hosted/ast.jou
Akuli: Actually I'll make it `name` so that I don't have `function_or_method_name_location`

self_hosted/runs_wrong.txt
Akuli: Because it uses a `return` statement without a return value.

self_hosted/runs_wrong.txt
Akuli: Because we don't have local variables yet, and it is attempting to return a local variable. I will change the test to return an integer constant instead.

PR #285, opened 2023-03-02
Fix compiler crash when trying to use the return value of a '-> void' method

Akuli:
There was already a nice error message for functions. I changed it to do the same thing for methods.


tests/other_errors/using_void_method.jou
```diff
@@ -0,0 +1,5 @@
+def f() -> void:
```
Akuli, line 1: The test file names are backwards.

PR #295, opened 2023-03-04
Test the produced zip files in GitHub Actions

Akuli:
I changed how Windows builds work in #293, and the generated `.zip` files are no longer tested: if they don't work for whatever reason, I probably wouldn't notice it until a user complains. This PR adds a test to improve the situation.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -61,6 +61,24 @@ jobs:
       - run: cd "test dir" && source activate && ./runtests.sh --verbose
         shell: bash
 
+  test-zip:
+    needs: build
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          path: repo
+      - uses: actions/download-artifact@v3
+        with:
+          name: windows-zip
+      - run: unzip jou.zip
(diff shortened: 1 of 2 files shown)


.github/workflows/windows.yml
```diff
@@ -61,6 +61,24 @@ jobs:
       - run: cd "test dir" && source activate && ./runtests.sh --verbose
         shell: bash
 
+  test-zip:
+    needs: build
```
Akuli, line 65:
```suggestion
    needs: build-zip
```

.github/workflows/windows.yml
```diff
@@ -61,6 +61,24 @@ jobs:
...
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          path: repo
+      - uses: actions/download-artifact@v3
+        with:
+          name: windows-zip
+      - run: unzip jou.zip
+      - run: cp -r repo/tests repo/runtests.sh jou
```
Akuli, line 75:
```suggestion
      - run: cp -r repo/tests repo/runtests.sh repo/activate jou
```

PR #296, opened 2023-03-04
Make the Windows zip files smaller

### LINKED ISSUE #294: Windows zips are huge
Previously a Jou zip file was 37MB. It will be 300MB now that #293 is merged. This is because it now contains a full copy of mingw.

I'm working on figuring out what can safely be deleted from mingw so that the linker still works. We really only need a linker from mingw.

Akuli:
Fixes #294 

The new zip files are about 93MB, so still pretty big but much better.


.github/workflows/windows.yml
```diff
@@ -22,22 +22,43 @@ jobs:
         shell: bash
       - run: source activate && mingw32-make
         shell: bash
-      - run: mkdir jou
-      - name: Copy files
+      # We don't need to copy all of mingw64. We only need the linker.
+      # The less we copy, the smaller the resulting zip becomes.
+      # What is needed was figured out with a script that deletes files one by one.
```
Akuli, line 27: This was how I approached it initially, but what we have now makes sense without knowing about the script:
```suggestion
```

.github/workflows/windows.yml
```diff
@@ -22,22 +22,43 @@ jobs:
...
+      - name: Copy some parts of mingw64 to jou/mingw64
+        shell: bash
+        run: |
+          # .a files compress well, so copy them all.
+          # CRT (C RunTime) files are needed because the linker implicitly adds them to every executable.
+          for file in $(find mingw64 -name '*.a') $(find mingw64 -name 'crt*.o'); do
```
Akuli, line 33:
```suggestion
          for file in $(find mingw64 -name '*.a') $(find mingw64 -name 'crt*.o') mingw64/version_info.txt; do
```

.github/workflows/windows.yml
```diff
@@ -22,22 +22,42 @@ jobs:
...
         shell: bash
-      - run: mkdir jou
-      - name: Copy files
+      # We don't need to copy all of mingw64. We only need the linker.
+      # The less we copy, the smaller the resulting zip becomes.
+      - name: Copy some parts of mingw64 to jou/mingw64
+        shell: bash
+        run: |
+          # .a files compress well, so copy them all.
+          # CRT (C RunTime) files are needed because the linker implicitly adds them to every executable.
```
Akuli, line 31:
```suggestion
          # CRT (C RunTime) files are needed because the linker implicitly adds them to every executable.
          # version_info.txt is only for curious humans.
```

PR #297, opened 2023-03-04
Speed up copying DLLs

Akuli:
This relatively simple step in the build on GitHub Actions currently takes 3min 30sec. Hopefully it is faster with this PR.


Akuli:
Now 50sec.

PR #298, opened 2023-03-04
Let's try gcc for linking on Windows

Akuli:
May further improve #294

### DIFF
--- a/src/run.c
+++ b/src/run.c
@@ -62,9 +62,10 @@ static void run_linker(const char *objpath, const char *exepath, const CommandLi
 
     char *command;
 #ifdef _WIN32
-    // Assume mingw with clang has been downloaded with windows_setup.sh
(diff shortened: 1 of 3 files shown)


src/run.c
```diff
@@ -62,9 +62,10 @@ static void run_linker(const char *objpath, const char *exepath, const CommandLi
 
     char *command;
 #ifdef _WIN32
-    // Assume mingw with clang has been downloaded with windows_setup.sh
+    // Assume mingw with clang has been downloaded with windows_setup.sh.
+    // Could also use clang, but but gcc has less dependencies so we can make the Windows zips smaller.
```
Akuli, line 66:
```suggestion
    // Could also use clang, but gcc has less dependencies so we can make the Windows zips smaller.
```

Akuli:
The zip file is now 37MB, so same as before.

PR #309, opened 2023-03-07
Emit multiple .o files and link them

Akuli:
This doesn't affect total compilation time much, but lets me see what object file takes the longest time to output. It turns out to be the slowest part of compiling.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -13,22 +13,6 @@
 #include <sys/stat.h>
 #include <llvm-c/TargetMachine.h>
 
-static void compile_to_object_file(LLVMModuleRef module, const char *path, const CommandLineFlags *flags)
-{
-    if (flags->verbose)
-        printf("Emitting object file \"%s\"\n", path);
-
-    char *tmppath = strdup(path);
-    char *error = NULL;
-    if (LLVMTargetMachineEmitToFile(get_target()->target_machine_ref, module, tmppath, LLVMObjectFile, &error)) {
-        assert(error);
-        fprintf(stderr, "failed to emit object file \"%s\": %s\n", path, error);
-        exit(1);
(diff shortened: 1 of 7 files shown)


src/output.c
```diff
@@ -49,111 +33,141 @@ static char *malloc_sprintf(const char *fmt, ...)
...
 {
     char *jou_exe = find_current_executable();
     const char *instdir = dirname(jou_exe);
 
-    char *linker_flags = malloc_sprintf("-lm %s", command_line_args.linker_flags ? command_line_args.linker_flags : "");
+    char *linker_flags;
+    if (command_line_args.linker_flags)
+        linker_flags = malloc_sprintf("-lm %s", command_line_args.linker_flags);
+    else
+        linker_flags = strdup("-lm");
```
Akuli, line 65: The "old" version was better

src/output.c
```diff
@@ -49,111 +33,141 @@ static char *malloc_sprintf(const char *fmt, ...)
...
-char *get_path_to_file_in_jou_compiled(LLVMModuleRef module, const char *filename)
+static char *get_path_to_file_in_jou_compiled(const char *filename)
 {
-    char *sourcepath = strdup(LLVMGetSourceFileName(module, (size_t[]){0}));
-
-    char *result = malloc(strlen(sourcepath) + strlen(filename) + 100);
-    char *tmp = strdup(sourcepath);
-    sprintf(result, "%s/jou_compiled", dirname(tmp));
-    free(tmp);
+    // Place it to current working directory. We need subdirectories based on the
```
Akuli, line 112: Maybe should instead be directory containing main Jou file, so you don't get race condition if multiple directory contain same name file

PR #311, opened 2023-03-08
Add size_t type

### LINKED ISSUE #304: 32-bit support
Jou does not have a `size_t` or `usize` type that would be 32-bit on some systems and 64-bit in others. This means that Tuomas can't use Jou on his raspberry pi. We experimented a bit, and apart from this problem, we got everything to work.

taahol:
Fixes #304


littlewhitecloud:
@taahol you also need add size_t type to the self-hosted compiler

Akuli:
This has a pile of merge conflicts and the approach is no longer what I want. Thanks for your work on Jou, even though this one didn't get merged :)

PR #312, opened 2023-03-10
Add a --small option to windows_setup.sh

### LINKED ISSUE #310: Why we need windows_setup.sh now?
Run jou on windows doesn't need windows_setup.sh before.
image
And it takes me a long time to download mingw64...

Akuli:
Fixes #310

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -27,6 +27,7 @@ jobs:
         with:
           ref: winlibs-small-zipfile
       - name: Download the full mingw64
+        # Same URL as in windows_setup.sh
         run: curl -L -o mingw64.zip https://github.com/brechtsanders/winlibs_mingw/releases/download/12.1.0-14.0.6-10.0.0-msvcrt-r3/winlibs-x86_64-posix-seh-gcc-12.1.0-llvm-14.0.6-mingw-w64msvcrt-10.0.0-r3.zip
       - name: Make sure that all file paths start with mingw64
         run: |
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -12,13 +12,50 @@ on:
   pull_request:
 
 jobs:
+  # The Jou repository includes a zipped copy of mingw64 that doesn't include some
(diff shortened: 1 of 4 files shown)


.github/workflows/windows.yml
```diff
@@ -12,17 +12,58 @@ on:
   pull_request:
 
 jobs:
+  # The Jou repository includes a zipped copy of mingw64 that doesn't include some
+  # large files needed for developing Jou. People with slow internet need it.
```
Akuli, line 16:
```suggestion
  # large files unnecessary for developing Jou. People with slow internet need it.
```

littlewhitecloud:
That's great!

.github/workflows/windows.yml
```diff
@@ -12,17 +12,69 @@ on:
   pull_request:
 
 jobs:
+  # The Jou repository includes a zipped copy of mingw64 that doesn't include some
+  # large files unnecessary for developing Jou. People with slow internet need it.
```
Akuli, line 16:
```suggestion
  # I created a zip file that contains mingw64, but with some large files deleted.
  # These large files are most of the time unnecessary for developing Jou.
  # People with slow internet need the smaller zip file.
```

.github/workflows/windows.yml
```diff
@@ -12,17 +12,69 @@ on:
...
+            exit 1
+          fi
+
   build-zip:
     # Even though jou.exe runs on windows, it is compiled on linux.
     # This is by far the easiest way to compile for Windows that I know of.
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@v3
+      # Can't use --small, because it is only for development and doesn't include GNU linker.
```
Akuli, line 72: This doesn't make sense to me. The compiler always invokes gcc.

Akuli:
@littlewhitecloud With `./windows_setup.sh --small`, the mingw download is about 60MB instead of 260MB. So with your internet connection, downloading it is going to take about 55 minutes instead of 4 hours. Not great, but not anywhere near as bad as before :)

PR #314, opened 2023-03-10
compare_compilers.sh improvements

Akuli:
- Fix CRLF problems on Linux (and on Windows, if your git config doesn't have autocrlf enabled)
- Complain if error list files mention nonexistent files (also works with `--fix`)
- Move `set -e` where it belongs, so that errors in argument handling will not be ignored

### DIFF
--- a/compare_compilers.sh
+++ b/compare_compilers.sh
@@ -93,12 +93,6 @@ for action in tokenize parse run; do
     (./jou $flag $file || true) &> tmp/compare_compilers/compiler_written_in_c.txt
     (./self_hosted_compiler $flag $file || true) &> tmp/compare_compilers/self_hosted.txt
 
-    if grep -q $'\r' $error_list_file; then
-        newline=crlf
-    else
-        newline=lf
-    fi
-
     if grep -qxF $file <(cat $error_list_file | tr -d '\r'); then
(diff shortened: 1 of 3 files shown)


compare_compilers.sh
```diff
@@ -54,7 +93,7 @@ for action in tokenize parse run; do
     (./jou $flag $file || true) &> tmp/compare_compilers/compiler_written_in_c.txt
     (./self_hosted_compiler $flag $file || true) &> tmp/compare_compilers/self_hosted.txt
 
-    if grep -q '\r' $error_list_file; then
+    if grep -q $'\r' $error_list_file; then
         newline=crlf
     else
         newline=lf
```
Akuli, line 99: no longer need

compare_compilers.sh
```diff
@@ -29,11 +31,48 @@ else
...
 mkdir -vp tmp/compare_compilers
 
+function append_line()
+{
+    local file="$1"
+    local line="$2"
+    echo "  Adding $line to $file"
+
+    if [ grep -q $'\r' $error_list_file ]; then
+        # CRLF line endings (Windows, but also depends on git settings)
```
Akuli, line 44:
```suggestion
        # CRLF line endings (likely Windows, but depends on git settings)
```

PR #318, opened 2023-03-11
Small code cleanup: combine related things

Akuli:
E.g. function declarations and function definitions now use the same data structure, and are distinguished by whether the body is empty.

This seems a bit weird at first, but IMO results in less copy/pasta and more readable code.

### DIFF
--- a/self_hosted/ast.jou
+++ b/self_hosted/ast.jou
@@ -298,7 +298,7 @@ class AstStatement:
     if_statement: AstIfStatement
     while_loop: AstConditionAndBody
     for_loop: AstForLoop
-    return_value: AstExpression*     # AstStatementKind::ReturnWithValue (can be NULL)
+    return_value: AstExpression*     # AstStatementKind::Return (can be NULL)
     assignment: AstAssignment
     var_declaration: AstNameTypeValue  # AstStatementKind::DeclareLocalVar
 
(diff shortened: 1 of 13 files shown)


self_hosted/ast.jou
```diff
@@ -299,7 +298,7 @@ class AstStatement:
     if_statement: AstIfStatement
     while_loop: AstConditionAndBody
     for_loop: AstForLoop
-    return_value: AstExpression     # AstStatementKind::ReturnWithValue
+    return_value: AstExpression*     # AstStatementKind::ReturnWithValue (can be NULL)
```
Akuli, line 301:
```suggestion
    return_value: AstExpression*     # AstStatementKind::Return (can be NULL)
```

src/print.c
```diff
@@ -300,12 +300,13 @@ static void print_ast_statement(const AstStatement *stmt, struct TreePrinter tp)
...
             break;
-        case AST_STMT_RETURN_VALUE:
-            printf("return a value\n");
-            print_ast_expression(&stmt->data.expression, print_tree_prefix(tp, true));
-            break;
-        case AST_STMT_RETURN_WITHOUT_VALUE:
-            printf("return without a value\n");
+        case AST_STMT_RETURN:
+            if (stmt->data.returnvalue) {
+                printf("return a value\n");
```
Akuli, line 305: Might as well just print "return". It's obvious from the output whether a return value is present or not.

PR #322, opened 2023-03-11
compare_compilers.sh improvements

Akuli:
- Accept `--fix` anywhere, not only as the first argument
- Don't run `make` twice. It is enough to compile the self-hosted compiler, that will also compiler the compiler written in C if necessary.
- Print some output lines with color. Don't color everything, so you can spot the unusual/colorful lines easily.
- ~Fix CRLF detection in `append_line` function~

### DIFF
--- a/compare_compilers.sh
+++ b/compare_compilers.sh
@@ -38,7 +38,7 @@ function append_line()
     local line="$2"
     echo -e "  ${YELLOW}Adding $line to $file${RESET}"
 
-    if [ grep -q $'\r' $error_list_file ]; then
+    if grep -q $'\r' $error_list_file; then
         # CRLF line endings (likely Windows, but depends on git settings)
         printf "%s\r\n" "$line" >> "$file"
     else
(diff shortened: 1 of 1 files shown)


compare_compilers.sh
```diff
@@ -6,39 +6,39 @@
...
+    else
+        files+=($arg)
+    fi
+done
 
-if [ $# = 0 ]; then
+if [ ${#files[@]} = 0 ]; then
     files=$(find stdlib examples tests -name '*.jou' | sort)
 else
     files=$*
```
Akuli, line 26:
```suggestion
```

compare_compilers.sh
```diff
@@ -52,7 +50,7 @@ function remove_line()
 {
     local file="$1"
     local line="$2"
-    echo "  Deleting $line from $file"
+    echo -e "  ${GREEN}Deleting $line from $file${RESET}"
```
Akuli, line 53: For consistency with the previous echo

```suggestion
    echo -e "  ${YELLOW}Deleting $line from $file${RESET}"
```

PR #324, opened 2023-03-11
More work on self hosted parser

Akuli:
It can now parse all files that don't use classes or enums or global variables, I think.


self_hosted/parser.jou
```diff
@@ -159,9 +196,22 @@ def parse_elementary_expression(tokens: Token**) -> AstExpression:
...
         expr.bool_value = False
         ++*tokens
-    elif (*tokens)->kind == TokenKind::Name and (&(*tokens)[1])->is_operator("("):
+    elif (*tokens)->is_keyword("NULL"):
+        expr.kind = AstExpressionKind::Null
+        ++*tokens
+    elif (*tokens)->kind == TokenKind::Name and (*tokens)[1].is_operator("("):
         expr.kind = AstExpressionKind::FunctionCall
         expr.call = parse_call(tokens, "(", ")")
+    elif (*tokens)->kind == TokenKind::Name and (*tokens)[1].is_operator("::") and (*tokens)[2].kind == TokenKind::Name:
```
Akuli, line 205: This could be more readable with nested if statements, because now many of the `elif` cases are only about `(*tokens)->kind == TokenKind::Name`.

PR #327, opened 2023-03-12
Union

### LINKED ISSUE #117: enums and unions
Currently jou does not have enums or unions.

I'm not yet sure if I want to separate `enum` and `union` like in C, or combine them into just one thing like rust's `enum`. Keeping them separate is less magical but also a lot less convenient.

### LINKED ISSUE #319: Compiling the self-hosted compiler is really slow
Emitting `parser.o` (the object file for `parser.jou`) takes about 1 minute, which is surprising because `parser.jou` isn't that complicated and the corresponding C code compiles in less than a second.

I think the reason is that the optimized LLVM IR looks very repetitive. It seems to initialize and copy large structs with one instruction for each struct element instead of using a `memcpy`.

Akuli:
Fixes #319 
Fixes #117

### DIFF
--- a/src/codegen.c
+++ b/src/codegen.c
@@ -30,6 +30,10 @@ Because the align is always a power of two, the memory will be suitably aligned 
 */
 static LLVMTypeRef codegen_union_type(const LLVMTypeRef *types, int ntypes)
 {
+    // Make the IR easier to read when unions aren't actually used
+    if (ntypes == 1)
+        return types[0];
+
     for (int i = 0; i < ntypes; i++) {
         assert(size(types[i]) > 0);
         assert(align(types[i])==1 || align(types[i])==2 || align(types[i])==4 || align(types[i])==8);
@@ -325,7 +329,7 @@ static void codegen_instruction(const struct State *st, const CfInstruction *ins
             }
             break;
         case CF_CONSTANT: setdest(codegen_constant(st, &ins->data.constant)); break;
-        case CF_SIZEOF: setdest(LLVMConstInt(LLVMInt64Type(), LLVMStoreSizeOfType(get_target()->target_data_ref, codegen_type(ins->data.type)), false)); break;
+        case CF_SIZEOF: setdest(LLVMConstInt(LLVMInt64Type(), size(codegen_type(ins->data.type)), false)); break;
         case CF_ADDRESS_OF_LOCAL_VAR: setdest(get_pointer_to_local_var(st, ins->operands[0])); break;
         case CF_ADDRESS_OF_GLOBAL_VAR: setdest(LLVMGetNamedGlobal(st->module, ins->data.globalname)); break;
         case CF_PTR_LOAD: setdest(LLVMBuildLoad(st->builder, getop(0), "ptr_load")); break;
@@ -346,8 +350,8 @@ static void codegen_instruction(const struct State *st, const CfInstruction *ins
 
                 LLVMValueRef val = LLVMBuildStructGEP2(st->builder, codegen_type(classtype), getop(0), f->union_id, ins->data.fieldname);
                 // This cast is needed in two cases:
-                //  * All pointers are codegenned as i8* so we can do self-referencing classes.
-                //  * It is how union types work.
+                //  * All pointers are i8* in structs so we can do self-referencing classes.
+                //  * This is how unions work.
                 val = LLVMBuildBitCast(st->builder, val, LLVMPointerType(codegen_type(f->type),0), "struct_member_cast");
                 setdest(val);
             }
--- a/src/codegen.c
+++ b/src/codegen.c
@@ -45,9 +45,13 @@ static LLVMTypeRef codegen_union_type(const LLVMTypeRef *types, int ntypes)
     for (int i = 0; i < ntypes; i++)
         sizeneeded = max(sizeneeded, size(types[i]));
 
+    // TODO: Figure out why this doesn't actually work, and causes the self-hosted compiler to crash / valgrind badly.
+#if 0
     unsigned count = (sizeneeded + size(chosen) - 1)/size(chosen);  // ceil division
     assert(count > 0);
     return LLVMArrayType(chosen, count);
+#endif
+    return LLVMArrayType(LLVMInt8Type(), sizeneeded);
 }
 
 static LLVMTypeRef codegen_type(const Type *type)
--- a/src/codegen.c
+++ b/src/codegen.c
@@ -1,12 +1,18 @@
 #include <assert.h>
 #include <stdbool.h>
 #include <stdio.h>
+#include <limits.h>
 #include <stdlib.h>
 #include <string.h>
 #include <llvm-c/Core.h>
 #include <llvm-c/Types.h>
 #include "jou_compiler.h"
 #include "util.h"
+
+static unsigned long long sizeof_llvm_type(LLVMTypeRef t)
+{
+    return LLVMStoreSizeOfType(get_target()->target_data_ref, t);
+}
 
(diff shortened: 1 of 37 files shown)


src/codegen.c
```diff
@@ -29,22 +75,36 @@ static LLVMTypeRef codegen_type(const Type *type)
         return LLVMIntType(type->data.width_in_bits);
     case TYPE_BOOL:
         return LLVMInt1Type();
-    case TYPE_OPAQUE_CLASS:
+    case TYPE_OPAQUE:
```
Akuli, line 78: Does not need renaming.

src/codegen.c
```diff
@@ -1,13 +1,59 @@
...
+https://mapping-high-level-constructs-to-llvm-ir.readthedocs.io/en/latest/basic-constructs/unions.html
+
+My first idea was to use an array of bytes that is big enough to fit anything.
+However, that might not be aligned properly.
+
+Instead, we choose the member type that has the biggest align, and make an array of it.
+Because the align is always a power of two, the memory will be suitably aligned for all member types.
+*/
+static LLVMTypeRef codegen_union_type(const LLVMTypeRef *types, int ntypes)
+{
```
Akuli, line 32:
```suggestion
{
    // Make the IR easier to read when unions aren't actually used
    if (ntypes == 1)
        return types[0];

```

src/codegen.c
```diff
@@ -265,7 +325,7 @@ static void codegen_instruction(const struct State *st, const CfInstruction *ins
             }
             break;
         case CF_CONSTANT: setdest(codegen_constant(st, &ins->data.constant)); break;
-        case CF_SIZEOF: setdest(LLVMSizeOf(codegen_type(ins->data.type))); break;
+        case CF_SIZEOF: setdest(LLVMConstInt(LLVMInt64Type(), LLVMStoreSizeOfType(get_target()->target_data_ref, codegen_type(ins->data.type)), false)); break;
```
Akuli, line 328:
```suggestion
        case CF_SIZEOF: setdest(LLVMConstInt(LLVMInt64Type(), size(codegen_type(ins->data.type)), false)); break;
```

src/codegen.c
```diff
@@ -281,17 +341,14 @@ static void codegen_instruction(const struct State *st, const CfInstruction *ins
...
-                }
 
-                LLVMValueRef val = LLVMBuildStructGEP2(st->builder, codegen_type(classtype), getop(0), i, ins->data.fieldname);
-                if (f->type->kind == TYPE_POINTER) {
-                    // We lied to LLVM that the struct member is i8*, so that we can do self-referencing types
-                    val = LLVMBuildBitCast(st->builder, val, LLVMPointerType(codegen_type(f->type),0), "struct_member_i8_hack");
-                }
+                LLVMValueRef val = LLVMBuildStructGEP2(st->builder, codegen_type(classtype), getop(0), f->union_id, ins->data.fieldname);
+                // This cast is needed in two cases:
+                //  * All pointers are codegenned as i8* so we can do self-referencing classes.
```
Akuli, line 349:
```suggestion
                //  * All pointers are i8* in structs so we can do self-referencing classes.
```

src/codegen.c
```diff
@@ -281,17 +341,14 @@ static void codegen_instruction(const struct State *st, const CfInstruction *ins
...
+                //  * It is how union types work.
```
Akuli, line 350:
```suggestion
                //  * This is how unions work.
```

src/free.c
```diff
@@ -203,14 +203,22 @@ void free_ast(AstToplevelNode *topnodelist)
...
+                    for (AstNameTypeValue *ntv = m->data.unionfields.ptr; ntv < End(m->data.unionfields); ntv++)
+                        free_name_type_value(ntv);
+                    free(m->data.unionfields.ptr);
+                    break;
+                case AST_CLASSMEMBER_METHOD:
+                    free_ast_signature(&m->data.method.signature);
+                    free_ast_body(&m->data.method.body);
+                    break;
+                }
             }
```
Akuli, line 221:
```suggestion
            }
            free(t->data.classdef.members.ptr);
```

src/jou_compiler.h
```diff
@@ -356,8 +373,8 @@ struct Type {
         TYPE_POINTER,
         TYPE_VOID_POINTER,
         TYPE_ARRAY,
+        TYPE_OPAQUE,  // unknown for now what exactly it is, will become something else later
         TYPE_CLASS,
-        TYPE_OPAQUE_CLASS,  // struct with unknown members
```
Akuli, line 360: This part should be mostyl reverted

src/jou_compiler.h
```diff
@@ -395,7 +412,7 @@ const Type *get_integer_type(int size_in_bits, bool is_signed);
 const Type *get_pointer_type(const Type *t);  // result lives as long as t
 const Type *get_array_type(const Type *t, int len);  // result lives as long as t
 const Type *type_of_constant(const Constant *c);
-Type *create_opaque_struct(const char *name);
+Type *create_opaque_type(const char *name);
```
Akuli, line 415: Should be reverrted.

src/parse.c
```diff
@@ -212,15 +212,15 @@ static AstCall parse_call(const Token **tokens, char openparen, char closeparen,
 
     while (!is_operator(*tokens, (char[]){closeparen,'\0'})) {
         if (args_are_named) {
-            // This code is only for structs, because there are no named function arguments.
+            // This code is only for classes, because there are no named function arguments.
```
Akuli, line 215:
```suggestion
            // This code is only for instantiating classes, because there are no named function arguments.
```

src/parse.c
```diff
@@ -810,6 +810,32 @@ static AstFunction parse_funcdef(const Token **tokens, bool is_method)
     return funcdef;
 }
 
+static void check_class_for_duplicate_names(const AstClassDef *classdef)
+{
+    List(struct MemberInfo { const char *name; const char *kindstr; Location location; }) infos = {0};
```
Akuli, line 815: Maybe would make more sense with `kindstr` and `name` order swapped.

src/typecheck.c
```diff
@@ -748,16 +772,14 @@ static const Type *typecheck_function_or_method_call(FileTypes *ft, const AstCal
...
 
-static const Type *typecheck_class_field(
-    const Type *classtype, const char *fieldname, Location location)
+static const struct ClassField *typecheck_class_field(
+    const Type *type, const char *fieldname, Location location)
 {
-    assert(classtype->kind == TYPE_CLASS);
-
-    for (struct ClassField *f = classtype->data.classdata.fields.ptr; f < End(classtype->data.classdata.fields); f++)
+    assert(type->kind == TYPE_CLASS);
```
Akuli, line 778: Might as well rename it back to classtype

src/typecheck.c
```diff
@@ -914,9 +949,9 @@ static ExpressionTypes *typecheck_expression(FileTypes *ft, const AstExpression
         if (temptype->kind != TYPE_CLASS)
             fail_with_error(
                 expr->location,
-                "left side of the '.' operator must be a class, not %s",
+                "left side of the '.' operator must be a class instance, not %s",
```
Akuli, line 952: "class instance" just sounds weird. Could probably grep for it, and replace it with "instance of a class"

src/types.c
```diff
@@ -165,10 +165,10 @@ const Type *type_of_constant(const Constant *c)
     assert(0);
 }
 
-Type *create_opaque_struct(const char *name)
+Type *create_opaque_type(const char *name)
```
Akuli, line 168: No need to rename this

PR #328, opened 2023-03-12
Change union codegen to be aligned properly

Akuli:
~Similar to how clang works, at least according to https://stackoverflow.com/a/12757239~

The previous solution of just using i8 array won't work on platforms that need stuff to be aligned. On x86_64 you don't need to align things, so that's why it worked, but it could cause problems if I port Jou to a different platform.

I changed it to array of i64 (of appropriate size), so it is always aligned as needed. It can be larger than necessary though, if your union is very small otherwise.

### DIFF
--- a/src/codegen.c
+++ b/src/codegen.c
@@ -18,7 +18,7 @@ My first idea was to use an array of bytes that is big enough to fit anything.
 However, that might not be aligned properly.
 
 Then I tried choosing the member type that has the biggest align, and making a large enough array of it.
-Because the align is always a power of two, the memory was be suitably aligned for all member types.
+Because the align is always a power of two, the memory will be suitably aligned for all member types.
 But it didn't work for some reason I still don't understand.
 
 Then I figured out how clang does it and did it the same way.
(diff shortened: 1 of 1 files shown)


src/codegen.c
```diff
@@ -10,52 +10,38 @@
...
 LLVM doesn't have a built-in union type, and you're supposed to abuse other types for that:
 https://mapping-high-level-constructs-to-llvm-ir.readthedocs.io/en/latest/basic-constructs/unions.html
 
 My first idea was to use an array of bytes that is big enough to fit anything.
 However, that might not be aligned properly.
 
-Instead, we choose the member type that has the biggest align, and make an array of it.
-Because the align is always a power of two, the memory will be suitably aligned for all member types.
+Then I tried choosing the member type that has the biggest align, and making a large enough array of it.
+Because the align is always a power of two, the memory was be suitably aligned for all member types.
```
Akuli, line 21:
```suggestion
Because the align is always a power of two, the memory will be suitably aligned for all member types.
```

src/codegen.c
```diff
@@ -10,52 +10,38 @@
...
-
-    // TODO: Figure out why this doesn't actually work, and causes the self-hosted compiler to crash / valgrind badly.
-#if 0
-    unsigned count = (sizeneeded + size(chosen) - 1)/size(chosen);  // ceil division
-    assert(count > 0);
-    return LLVMArrayType(chosen, count);
-#endif
-    return LLVMArrayType(LLVMInt8Type(), sizeneeded);
+        // If this assert fails, you need to figure out which of the size functions should be used.
+        // I don't know what their difference is.
```
Akuli, line 40:
```suggestion
        // I don't know what their difference is.
        // And if you need the alignment, there's 3 different functions for that...
```

PR #334, opened 2023-03-13
Run release once per day

Akuli:
Right now it makes a release for almost every PR which seems unnecessary.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@
 #
 on:
   schedule:
-    - cron: '0 */3 * * *'
+    - cron: '0 4 * * *'
   workflow_dispatch:  # Can also be triggered manually from github UI
 
 jobs:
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -156,14 +156,15 @@ To make using Jou on Windows equally easy, or at least as painless as possible,
 users can download a zip file that contains everything needed to run Jou.
 See the README for instructions to install Jou from a zip file.
 
-A zip file is built and released in GitHub Actions every 3 hours,
+A zip file is built and released in GitHub Actions every morning,
```
Akuli, line 159:
```suggestion
A zip file is built and released in GitHub Actions every morning at 4AM UTC,
```

.github/workflows/release.yml
Akuli, line 1: This is a test. Github apparently lets you comment an entire file, so I'm trying that.

PR #337, opened 2023-03-14
Improve ensure_can_take_address

Akuli:
The purpose of `ensure_can_take_address()` is to check that making a pointer to something (as in `&foo`) makes sense. But unlike you would expect, it isn't used just in the `&foo` syntax. Several other things, such as `foo = bar` and `&foo.bar`, also need a pointer to `foo`. The `ensure_can_take_address()` function handles them all.

This means that it is important for `ensure_can_take_address()` to generate good error messages. The error message logic is also a bit complicated, and previously behaved wrong in some corner cases: `Foo{}.x++` resulted in an error message saying you are trying to increment a newly created instance. It now says that you are trying to increment a *field* of a newly created instance.

### DIFF
--- a/src/typecheck.c
+++ b/src/typecheck.c
@@ -611,6 +611,8 @@ static void ensure_can_take_address(const AstExpression *expr, const char *errms
     case AST_EXPR_GET_FIELD:
         // &foo.bar = &foo + offset
         {
+            // Turn "cannot assign to %s" into "cannot assign to a field of %s".
+            // This assumes that errmsg_template is relatively simple, i.e. it only contains one %s somewhere.
             char *newtemplate = malloc(strlen(errmsg_template) + 100);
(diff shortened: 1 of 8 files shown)


src/typecheck.c
```diff
@@ -610,7 +610,12 @@ static void ensure_can_take_address(const AstExpression *expr, const char *errms
         break;
     case AST_EXPR_GET_FIELD:
         // &foo.bar = &foo + offset
-        ensure_can_take_address(&expr->data.operands[0], errmsg_template);
+        {
```
Akuli, line 613:
```suggestion
        {
            // Turn "cannot assign to %s" into "cannot assign to a field of %s".
            // This assumes that errmsg_template is relatively simple, i.e. it only contains one %s somewhere.
```

PR #338, opened 2023-03-14
Self-hosted compiler: classes

Akuli:


self_hosted/typecheck.jou
```diff
@@ -489,6 +642,84 @@ class Stage3TypeChecker:
...
+                t,
+                &instantiation->field_names[i][0],
+            )
+            self->do_expression_and_implicit_cast(
+                &instantiation->field_values[i],
+                specified_fields[i]->type,
+                &message[0],
+            )
+
+        # TODO: check union_id
```
Akuli, line 719: might as well do

src/typecheck.c
```diff
@@ -796,8 +817,8 @@ static const Type *typecheck_struct_init(FileTypes *ft, const AstCall *call, Loc
 
     if (t->kind != TYPE_CLASS) {
         fail_with_error(location,
-            "the %s{...} syntax is only for classes, not for type %s",
-            t->name, t->name);
+            "the %s{...} syntax is only for classes, but %s is %s",
```
Akuli, line 820: Presumably tests need to be changed for this.

PR #339, opened 2023-03-15
Add implicit cast from array to pointer

### LINKED ISSUE #329: &foo[0] is annoying
If I have an array `foo: byte[100]`, I need to do `&foo[0]` to pass it to a function like `strcmp()`, `strcpy()`, `snprintf()`, `printf()`, ...

This means that Jou code contains a LOT of `&foo[0]` compared to C. I don't like it. It would be better if the array automatically converted into a pointer to its first argument, just like in C.

Akuli:
Fixes #329


self_hosted/typecheck.jou
```diff
@@ -60,13 +60,13 @@ def fail_with_implicit_cast_error(location: Location, template: byte*, from: Typ
...
-            strcat(message, &from->name[0])
+            strcat(message, from->name)
         elif starts_with(template, "<to>"):
             template = &template[4]
-            strcat(message, &to->name[0])
+            strcat(message, to->name)
         else:
             s = [*template++, '\0']
-            strcat(message, &s[0])
+            strcat(message, s)
```
Akuli, line 69: Without the explicit `&` operator, it will probably feel weird to many people why you need to place the array to a temporary variable. For now I should ensure that the error message is clear .

src/typecheck.c
```diff
@@ -392,14 +397,21 @@ static void do_implicit_cast(
...
+{
+    assert(types->type->kind == TYPE_ARRAY);
+    do_implicit_cast(types, get_pointer_type(types->type->data.array.membertype), (Location){0}, NULL);
 }
 
 static void check_explicit_cast(const Type *from, const Type *to, Location location)
 {
     if (
         from != to  // TODO: should probably be error if it's the same type.
+        && !(from->kind == TYPE_ARRAY && to->kind == TYPE_POINTER && from->data.array.membertype == to->data.valuetype)
```
Akuli, line 414: Should test the explicit casting

src/typecheck.c
```diff
@@ -392,14 +397,21 @@ static void do_implicit_cast(
     if (errormsg_template != NULL && !can_cast_implicitly(from, to))
         fail_with_implicit_cast_error(location, errormsg_template, from, to);
 
-    assert(!types->implicit_cast_type);
     types->implicit_cast_type = to;
+    types->implicit_array_to_pointer_cast = (from->kind == TYPE_ARRAY && to->kind == TYPE_POINTER);
+}
+
+static void cast_array_to_pointer(ExpressionTypes *types)
```
Akuli, line 404: Presumably the indexing syntax `foo[bar]` should use this too.

tests/should_succeed/array.jou
```diff
@@ -33,7 +33,10 @@ def main() -> int:
     do_nothing(foo)
     printf("%d %d %d\n", foo[0], foo[1], foo[2])  # Output: 4 5 6
 
-    increment(&foo[0])  # Needs explicit &foo[0], doesn't convert implicitly like in C.
+    increment(&foo[0])
     printf("%d %d %d\n", foo[0], foo[1], foo[2])  # Output: 5 5 6
 
+    increment(foo)  # foo converts implicitly to &foo[0]
```
Akuli, line 39:
```suggestion
    increment(foo)  # foo casts implicitly to &foo[0]
```

PR #341, opened 2023-03-17
"Pass" syntax support

### LINKED ISSUE #320: pass statement
Sometimes it's useful to have a "do nothing" statement, like a `pass` statement in Python. In Jou I used this:

https://github.com/Akuli/jou/blob/ba844331ec27263a61180fa68c9029a63e2e5099/self_hosted/tokenizer.jou#L315-L317

And for a function that does nothing (and also returns nothing), you can use the `return` keyword:

https://github.com/Akuli/jou/blob/ba844331ec27263a61180fa68c9029a63e2e5099/tests/already_exists_error/func.jou#L1-L2

Both of these feel like workarounds and I think I should just add a `pass` statement.

littlewhitecloud:
Fixes #320 
Successed


Akuli:
Please add a test and also support this in the self-hosted compiler.

Akuli:
Closing and reopenning to re-run github actions. I hope I fixed it...

littlewhitecloud:
Now "pass" syntax like:
```python
import "stdlib/io.jou"

def main() -> int:
    a = 114514
    if (a == 114514):
        pass a # Must give somthing here (Don't know why)
    else:
        puts("wat?")
    return 0
```
But I hope it just like python:
```
if ...:
   pass
```

littlewhitecloud:
Code :
```
import "stdlib/io.jou"

def main() -> int:
    a = 114514
    if (a == 114514):
        pass
    else:
        puts("wat?")
    return 0
```
If you don't give anything to the "pass"
Compiler will say:
```
jou passtest.jou
compiler error in file "passtest.jou", line 6: expected an expression, got end of line
The process failed with status 1.
```

littlewhitecloud:
image
Failed run self-hosted compiler...
```
+./compare_compilers.sh: line 92: ./self_hosted_compiler: cannot execute binary file: Exec format error
  Error: Compilers behave differently when given "examples/fib.jou".
  Ideally the compilers would behave in the same way for all files, but we aren't there yet.
  To silence this error, add "examples/fib.jou" to self_hosted/tokenizes_wrong.txt (or run again with --fix).
```

littlewhitecloud:
```
--- tmp/compare_compilers/compiler_written_in_c.txt	2023-03-17 23:33:48.382277900 +0000
+++ tmp/compare_compilers/self_hosted.txt	2023-03-17 23:33:48.429153400 +0000
@@ -10,7 +10,7 @@
 | | |--- [line 5] get variable "a"
 | | `--- [line 5] 114514 (32-bit signed)
 | |--- body:
-| | `--- [line 6] pass

+| | `--- [line 6] break

 | `--- else body:
 |   `--- [line 8] expression statement
 |     `--- [line 8] call function "puts"
  Error: Compilers behave differently when given "tests/should_succeed/passtest.jou".
```
The "pass" syntax output has some error, Maybe I use wrong usage to output? (TreePrinter)

self_hosted/parser.jou
```diff
@@ -551,6 +551,9 @@ def parse_oneline_statement(tokens: Token**) -> AstStatement:
         ++*tokens
         result.kind = AstStatementKind::Assert
         result.expression = parse_expression(tokens)
+    elif (*tokens)->is_keyword("pass"):
+        ++*tokens
+        result.kind = AstStatementKind::Break
```
littlewhitecloud, line 556: 6

littlewhitecloud:
Throught we can use "pass" in the file now :)

self_hosted/ast.jou
```diff
@@ -482,6 +486,14 @@ class AstIfStatement:
         free(self->if_and_elifs)
         self->else_body.free()
 
+class AstPass:
```
Akuli, line 489: We don't need this class. It's fine to not have a `union` member for all different kinds of `AstStatement`s. There's no `AstBreak` or `AstContinue` classes either.

self_hosted/create_llvm_ir.jou
```diff
@@ -566,6 +569,8 @@ class AstToIR:
         elif ast->kind == AstStatementKind::Assert:
             condition = self->do_expression(&ast->expression)
             self->do_assert(condition)
+        elif ast->kind == AstStatementKind::Pass:
+            self->do_pass()
```
Akuli, line 573: Because we now have a `pass` keyword, we can just use it here:

```suggestion
            pass
```

You can then also delete the `do_pass()` method.

self_hosted/tokenizer.jou
```diff
@@ -22,7 +22,7 @@ def is_keyword(word: byte*) -> bool:
     #   - syntax documentation
     keywords = [
         "import", "def", "declare", "class", "union", "enum", "global",
-        "return", "if", "elif", "else", "while", "for", "break", "continue",
+        "return", "if", "elif", "else", "while", "for", "break", "continue", "pass",
```
Akuli, line 25: See the comment just above this code. You need to add the `pass` keyword to `doc/syntax-spec.md` where there is a list of all keywords

self_hosted/typecheck.jou
```diff
@@ -974,6 +974,9 @@ class Stage3TypeChecker:
             self->nested_loop_count--
             self->do_statement(statement->for_loop.incr)
 
+        elif statement->kind == AstStatementKind::Pass: # Do nothing here
+            assert True
```
Akuli, line 978:
```suggestion
            pass
```

self_hosted/tokenizer.jou
Akuli: Nevermind, it is there already. I didn't see it for some reason...

self_hosted/create_llvm_ir.jou
```diff
@@ -287,6 +287,9 @@ class AstToIR:
         printf("%s %d %s\n", lhs_type->name, op, rhs_type->name)
         assert False
 
+    def do_pass(self) -> void:
+        return
+
```
Akuli, line 292: This method is never used:

```suggestion
```

littlewhitecloud:
By the way, thr porcupine editor is very good :)

littlewhitecloud:
Thank you for checking out this pull request!

Akuli:
Thanks for working on my project! :)

PR #344, opened 2023-03-19
Handle asserts that are split across multiple lines

Akuli:


tests/other_errors/assert_fail_multiline.jou
```diff
@@ -0,0 +1,11 @@
+def main() -> int:
+    lol = True
+    wat = True
+    assert lol and wat
+    assert (  # Output: Assertion '(lol and not wat)' failed in file "tests/other_errors/assert_fail_multiline.jou", line 5.
```
Akuli, line 5:
```suggestion
    assert (  # Output: Assertion '(lol and not wat)' failed in file "tests/other_errors/assert_fail_multiline.jou", line 4.
```

Issue #244, opened 2023-02-25
No error for missing self parameter


Akuli:
```
struct Foo:
    def bar() -> void:
        return

def main() -> int:
    return 0
```

Expected result: error on line 2 because it is missing the `self` parameter

Actual result: no error

PR #345, opened 2023-03-20
Fix werid error

### LINKED ISSUE #275: Weird error message....
(Originally this issue contained an error that had to do with #244, but that is now a separate issue so I rewrote this issue description)

```python
class Foo:
    def bar(self, x: int) -> void:
        return

def main() -> int:
    f = Foo{}
    f.bar(1, 2)
    return 0
```

This fails with a confusing error message: `compiler error in file "asd.jou", line 7: method bar(self: Foo*, x: int) takes 2 arguments, but it was called with 2 arguments`

So the method expects 2 arguments (including self), but you give it 2 arguments (excluding self).

I think a better error message would be: `method bar(x: int) takes 1 argument, but it was called with 2 arguments`. Note that `self` is not mentioned in `bar(x: int)` and is also not counted into how many arguments the method wants.

littlewhitecloud:
fix #275 #244


tests/syntax_error/missing_self.jou
```diff
@@ -0,0 +1,8 @@
+class Foo:
+    def bar() -> void:
```
Akuli, line 2: Ideally the error message would be on this line. You can check it in the parser, maybe `parse_function_signature()`.

littlewhitecloud:
```
-compiler error in file "tests/syntax_error/missing_self.jou", line 2: error when defining method bar() bacause it has no self
+compiler error in file "tests/syntax_error/missing_self.jou", line 7: error when defining method bar() bacause it has no self
```
line 7
???

littlewhitecloud:
```
-compiler error in file "tests/other_errors/function_wrong_n_args.jou", line 4: function getchar() takes 0 arguments, but it was called with 1 argument
-Exit code: 1
+bash: line 1:  3861 Segmentation fault      (core dumped) ./jou -O3 tests/other_errors/function_wrong_n_args.jou
```
Oh, no, segmentation fault

src/typecheck.c
```diff
@@ -779,17 +779,26 @@ static const Type *typecheck_function_or_method_call(FileTypes *ft, const AstCal
...
-            "%s %s takes %d argument%s, but it was called with %d argument%s",
-            self_type ? "method" : "function",
-            sigstr,
-            sig->nargs,
-            sig->nargs==1?"":"s",
-            call->nargs,
-            call->nargs==1?"":"s"
-        );
+    if (n < sig->nargs || (n > sig->nargs && !sig->takes_varargs)) { // Check if it is a function in the class
+        if (self_type->kind != TYPE_CLASS) {
```
Akuli, line 783: It is segfaulting because `self_type` can be NULL here. It means that you have a function, not a method.

But as I said, typecheck is not the right place to detect a missing `self` argument. It should be done already when parsing.

Akuli:
> ```
> -compiler error in file "tests/syntax_error/missing_self.jou", line 2: error when defining method bar() bacause it has no self
> +compiler error in file "tests/syntax_error/missing_self.jou", line 7: error when defining method bar() bacause it has no self
> ```

This means that the error is reported on the wrong line. The compiler actually reports an error on line 7, but according to the test it should be reported on line 2.

littlewhitecloud:
my idea:
create a new flag named selfused, and if it doesn't true after the "while" loop and accept_self is true, I will give an parse_error

src/typecheck.c
littlewhitecloud: changed check in parse

littlewhitecloud:
Now the problem is: how can I get this function name for showing the error

Akuli:
You can use `result.name`.

src/parse.c
```diff
@@ -184,7 +184,15 @@ static AstSignature parse_function_signature(const Token **tokens, bool accept_s
         fail_with_parse_error(*tokens, "a '->'");
     }
     ++*tokens;
-
+	
+	if (!selfusedflag && accept_self) {
+        fail_with_error(
+            (*tokens)->location,
+            "error when defining %s bacause it has no self",
```
Akuli, line 191: Maybe like this:
```suggestion
            "missing self, should be def %s(self, ...)",
```

Akuli:
Now add the same error message in the self-hosted compiler :)

src/parse.c
```diff
@@ -184,7 +184,15 @@ static AstSignature parse_function_signature(const Token **tokens, bool accept_s
         fail_with_parse_error(*tokens, "a '->'");
     }
     ++*tokens;
-
+	
+	if (!used_self && accept_self) {
+        fail_with_error(
+            (*tokens)->location,
+            "missing self, should be 'def %s(self, ...)",
```
littlewhitecloud, line 191:
```suggestion
            "missing self, should be 'def %s(self, ...)'",
```

self_hosted/parser.jou
```diff
@@ -124,6 +126,9 @@ def parse_function_or_method_signature(tokens: Token**, is_method: bool) -> AstS
         (*tokens)->fail_expected_got("a '->'")
     ++*tokens
 
+    if not used_self and is_method:
+        fail((*tokens)->location, "missing self, should be 'def %s(self, ...)" % result.name)
```
Akuli, line 130: In the self-hosted compiler you need to use `sprintf()` or `snprintf()` to format error messages. For example:
```suggestion
        message: byte[300]
        snprintf(message, sizeof message, "missing self, should be 'def %s(self, ...)", result.name)
        fail((*tokens)->location, message)
```
Ideally the `fail()` function would do this for you, like `fail_with_error()` does in C, but it is not yet possible to define a printf-like function in Jou.

Akuli (approved):
Thanks again!

PR #348, opened 2023-03-20
Pointer comparisons

### LINKED ISSUE #279: cannot >= compare pointers
this doesn't work

```
    for p = &digits[strlen(digits) - 1]; p >= digits; p--:
```

this works

```
    for i = strlen(digits) - 1; i >= 0; i--:
```

I think both should work.

Akuli:
E.g. `&foo[0] < &foo[1]`. Fixes #279

### DIFF
--- a/src/tokenize.c
+++ b/src/tokenize.c
@@ -80,15 +80,15 @@ static void read_identifier_or_number(struct State *st, char firstbyte, char (*d
 
     assert(is_identifier_or_number_byte(firstbyte));
     (*dest)[destlen++] = firstbyte;
-    bool is_number = '0'<=firstbyte && firstbyte<='9';
 
     while(1) {
         char c = read_byte(st);
         if (is_identifier_or_number_byte(c)
-            || (is_number && (c=='.' || (c=='-' && (*dest)[destlen-1]=='e'))))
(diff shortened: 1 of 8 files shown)


src/typecheck.c
```diff
@@ -641,7 +642,7 @@ static const Type *check_binop(
     if (got_numbers && !got_integers)
         cast_type = (lhstypes->type == doubleType || rhstypes->type == doubleType) ? doubleType : floatType;
     if (got_pointers)
-        cast_type = voidPtrType;
+        cast_type = longType;
```
Akuli, line 645: Could be slightly safer to use unsigned long

```suggestion
        cast_type = get_integer_type(64, false);  // unsigned long
```

src/tokenize.c
```diff
@@ -80,15 +80,15 @@ static void read_identifier_or_number(struct State *st, char firstbyte, char (*d
 
     assert(is_identifier_or_number_byte(firstbyte));
     (*dest)[destlen++] = firstbyte;
+    bool is_number = '0'<=firstbyte && firstbyte<='9';
```
Akuli, line 83: Nice, but doesn't belong to this PR.

PR #356, opened 2023-03-25
.editorconfig

### LINKED ISSUE #355: Add .editorconfig to this repository
So that editors will automatically indent with 4 spaces and the like.

littlewhitecloud:
Improve #355

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,5 +4,5 @@ charset = utf-8
 trim_trailing_whitespace = true
 indent_size = 4
 
-[*.md]
-trim_trailing_whitespace = false
+[{Makefile,Makefile.*,tests/syntax_error/tab.jou}]
+indent_style = tab
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,3 +6,5 @@ indent_size = 4
 
 [{Makefile,Makefile.*,tests/syntax_error/tab.jou}]
 indent_style = tab
+
+[.github/workflows/*.yml]
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+[*]
+indent_style = space
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+indent_size = 4
+
+[*.md]
+trim_trailing_whitespace = false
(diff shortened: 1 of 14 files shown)


littlewhitecloud:
lf end
4 space
utf-8

.editorconfig
```diff
@@ -0,0 +1,9 @@
+[*]
+indent_style = space
+end_of_line = lf
```
Akuli, line 3: Git for Windows converts the files to CRLF automagically, so even though they are committed as LF, cloning the repository on Windows will actually give you CRLF.

```suggestion
```

.editorconfig
```diff
@@ -0,0 +1,9 @@
...
+charset = utf-8
+trim_trailing_whitespace = true
+indent_size = 4
+
+[*.md]
+trim_trailing_whitespace = false
```
Akuli, line 9: We don't need markdown to be special, but we do need to indent Makefiles with tabs. There's also one test file that is indented with a tab.

```suggestion
[{Makefile,Makefile.*,tests/syntax_error/tab.jou}]
indent_style = tab
```

Akuli:
Maybe also add to github actions some tool that checks whether files follow the editorconfig? It could go to `.github/workflows/linux.yml`. Searching "editorconfig validator" gives me https://github.com/editorconfig-checker/editorconfig-checker which seems useful.

littlewhitecloud:
I agree that.But I am not good at adding GitHub actions(lol)

Akuli:
We could copy/paste the example in https://github.com/editorconfig-checker/action-editorconfig-checker

littlewhitecloud:
let me try later

littlewhitecloud:
Oh no, too many errors 
I think we should add something like autofix

littlewhitecloud:
image
That's really strange.
typecheck.jou only has 1147 lines, but it say 1176 has trailing whitespace

Akuli:
The line numbers seem to be off for whatever reason, but I searched for spaces at ends of the lines and fixed the remaining ones.

littlewhitecloud:
Finally…

Akuli (approved):
Thanks for working on this!

PR #364, opened 2023-03-25
Self-hosted compiler: methods

Akuli:


self_hosted/create_llvm_ir.jou
```diff
@@ -133,6 +133,8 @@ class AstToIR:
         assert False
 
     def declare_function(self, sig: Signature*) -> LLVMValue*:
+        # FIXME: naming of methods
```
Akuli, line 136: Currently two methods of the same name in different classes will collide. Do we have a test about this?

self_hosted/typecheck.jou
```diff
@@ -196,9 +196,14 @@ class FileTypes:
                 return &self->all_functions[i]
         return NULL
 
-    def find_defined_function(self, name: byte*) -> FunctionOrMethodTypes*:
+    # If class_type is NULL, this finds a method
```
Akuli, line 199:
```suggestion
    # If class_type is NULL, this finds a function
```

tests/should_succeed/method.jou
```diff
@@ -8,11 +8,22 @@ class Foo:
         return self->x + self->y
 
     def increment_x(self) -> void:
-        self->x += 1
+        self->x++
+
+class Foo3:
+    # Different classes can have members with the same names.
```
Akuli, line 14:
```suggestion
    # Different classes can have methods with the same name.
```

PR #366, opened 2023-03-26
Convert string constants to arrays

### LINKED ISSUE #359: Initializing byte array with a string literal
https://github.com/Akuli/jou/blob/6f1c37dc2dc1fba055b08c29f8373618e9bf0613/self_hosted/parser.jou#L104-L108

This is a bit weird. Ideally you could just do `name = "self"`.

As a minimal example, this doesn't work:
```
foo: byte[100] = "hello"
```

## Solution 1: bidirectional type inference

Bidirectional type inference, aka type context, means that the type of a thing depends on how it is used. If the string `"hello"` is being assigned into an array of type `byte[100]`, its type would become `byte[100]` rather than `byte*`.

In general, I don't want to have bidirectional type inference in Jou, because it makes the language and the compilers more complicated. But just special-casing strings could be super simple.

I think I will go with this approach.

## Solution 2: `"hello"` has type `byte[6]`

If we add an implicit conversion that lets you assign `byte[6]` to `byte[100]` (presumably filling the end with zeros), and we change `"hello"` to be `byte[6]` rather than `byte*`, then `foo: byte[100] = "hello"` would work.

An advantage is that `sizeof "hello"` would produce 6 as expected, rather than 8 (size of a pointer on 64-bit computer).

A disadvantage is:

```
if something:
    foo = "asd"  # Type of variable blah becomes byte[4]
else:
    foo = "asd asd"  # Error, because this string doesn't fit in byte[4]
```

Another disadvantage is that you might want an error for assigning arrays of different sizes. For example, if you pass `int[3]` to a function expecting `int[5]`, you probably want an error rather than padding the `int[3]` array with two zeros.

Akuli:
Fixes #359

### DIFF
--- a/self_hosted/typecheck.jou
+++ b/self_hosted/typecheck.jou
@@ -464,7 +464,7 @@ def short_expression_description(expr: AstExpression*) -> byte[200]:
         or expr->kind == AstExpressionKind::Bool
         or expr->kind == AstExpressionKind::Null
     ):
-        strcpy(result, "a constant")
+        return "a constant"
     elif (
         expr->kind == AstExpressionKind::Negate
         or expr->kind == AstExpressionKind::Add
@@ -473,7 +473,7 @@ def short_expression_description(expr: AstExpression*) -> byte[200]:
         or expr->kind == AstExpressionKind::Divide
         or expr->kind == AstExpressionKind::Modulo
     ):
-        strcpy(result, "the result of a calculation")
+        return "the result of a calculation"
     elif (
         expr->kind == AstExpressionKind::Eq
         or expr->kind == AstExpressionKind::Ne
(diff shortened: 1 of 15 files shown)


self_hosted/typecheck.jou
```diff
@@ -149,6 +170,12 @@ class ExpressionTypes:
         assert self->original_type->kind == TypeKind::Array
         self->do_implicit_cast(self->original_type->array.item_type->get_pointer_type(), Location{}, NULL)
 
+    def cast_string_to_array(self, array_size: int) -> void:
```
Akuli, line 173: Is this ever used?

Akuli:
For some reason this makes compiling the self-hosted compiler about 6 times slower. I'm timing it with this command:

```
rm self_hosted_compiler && make jou && time make self_hosted_compiler
```

Akuli:
Now compiles in about same time as before.

src/build_cfg.c
```diff
@@ -158,6 +197,9 @@ static const LocalVariable *build_cast(
         return result;
     }
 
+    if (obj->type == get_pointer_type(byteType) && to->kind == TYPE_ARRAY && to->data.array.membertype == byteType)
+        return convert_string_to_array(st, obj, to, location);
```
Akuli, line 201: I believe this code never runs.

PR #371, opened 2023-03-27
fix weird error x2

### LINKED ISSUE #275: Weird error message....

littlewhitecloud:
fix #275

### DIFF
--- a/src/typecheck.c
+++ b/src/typecheck.c
@@ -795,16 +795,19 @@ static const Type *typecheck_function_or_method_call(FileTypes *ft, const AstCal
     }
 
     char *sigstr = signature_to_string(sig, false);
-
     int n = call->nargs + !!self_type;
+    int nargs = sig->nargs;
+    if (self_type) {
+		nargs--;
+	}
     if (n < sig->nargs || (n > sig->nargs && !sig->takes_varargs)) {
-        fail_with_error(
+		fail_with_error(
             location,
             "%s %s takes %d argument%s, but it was called with %d argument%s",
             self_type ? "method" : "function",
             sigstr,
-            sig->nargs,
-            sig->nargs==1?"":"s",
+            nargs,
+            nargs==1?"":"s",
             call->nargs,
             call->nargs==1?"":"s"
         );
--- a/tests/syntax_error/selft_take_arg.jou
+++ b/tests/syntax_error/selft_take_arg.jou
@@ -1,0 +1,8 @@
+class Foo:
+    def bar(self, x: int) -> void:
+        return
+
+def main() -> int:
+    f = Foo{}
+    f.bar(1, 2) # Error: method bar(self: Foo*, x: int) takes 1 argument, but it was called with 2 arguments
+    return 0


littlewhitecloud:
Oh, seems like self-hosted compiler solved this problem long ago.
But I think to show error, we still need to show ```self```

self_hosted/typecheck.jou
```diff
@@ -770,7 +770,7 @@ class Stage3TypeChecker:
                 )
             fail(call->location, message)
 
-        signature_string = signature->to_string(False, False)
+        signature_string = signature->to_string(True, False) # Throught we should still show self
```
littlewhitecloud, line 773:
```suggestion
        signature_string = signature->to_string(True, False)
```

tests/syntax_error/self_take_arg.jou
```diff
@@ -0,0 +1,8 @@
+class Foo:
+    def bar(self, x: int) -> void:
+        return
+
+def main() -> int:
+    f = Foo{}
+    f.bar(1, 2) # Error: method bar(self: Foo*, x: int) takes 1 argument, but it was called with 2 arguments
```
Akuli, line 7: I think it would be nicer if the error message didn't include `self: Foo*`. Instead of `bar(self: Foo*, x: int)` it would say `bar(x: int)`, because when calling a method, you only need to care about the `x`, not about `self`.

To do this, you can pass `False` in self-hosted compiler and add a similar parameter to `signature_to_string()` in the other compiler.

Akuli:
> I hate this!!!!

Now that we know our editorconfig file is somewhat accurate, we could delete the check in github actions (in a separate PR). I don't care much if a few C files have a few tab characters or some lines that contain unnecessary spaces.

tests/syntax_error/self_take_arg.jou
littlewhitecloud: okay

littlewhitecloud:
LOL, I just get tired of it...

src/typecheck.c
```diff
@@ -794,17 +794,20 @@ static const Type *typecheck_function_or_method_call(FileTypes *ft, const AstCal
...
     }
 
-    char *sigstr = signature_to_string(sig, false);
+    char *sigstr = signature_to_string(sig, false, false);
 
     int n = call->nargs + !!self_type;
+    int nargs = sig->nargs;
+    if (self_type)
+        nargs--;
     if (n < sig->nargs || (n > sig->nargs && !sig->takes_varargs)) {
```
Akuli, line 803: I think we can rewrite this line with `nargs` instead of `n`:
```suggestion
    if (call->nargs < nargs || (call->nargs > nargs && !sig->takes_varargs)) {
```
This way, we no longer need the `n` variable defined above.

littlewhitecloud:
okay

PR #377, opened 2023-04-01
More work on the stdlib

### LINKED ISSUE #376: types.jou?
I found a "TODO" in ast.jou
```python
# TODO: move to stdlib
declare isprint(b: int) -> int
```
I have created a pr(#227), but it failed, beacause it doesn't return anything...
```python
# Check if character is alphanumeric
declare isalnum(c: bytes) -> bool
# Check if character is alphabetic
declare isalpha(c: bytes) -> bool
# Check if character is a control character
declare iscntrl(c: bytes) -> bool
# Check if character is decimal digit
declare isdigit(c: int) -> bool
# Check if character has graphical representation
declare isgraph(c: bytes) -> bool
# Check if character is lowercase letter
declare islower(c: bytes) -> bool
# Check if character is printable 
declare isprint(c: bytes) -> bool
# Check if character is a punctuation character
declare ispunct(c: bytes) -> bool
# Check if character is a white-space
declare isspace(c: bytes) -> bool
# Check if character is uppercase letter
declare isupper(c: bytes) -> bool
# Check if character is hexadecimal digit
declare isxdigit(c: int) -> bool
# Convert uppercase letter to lowercase
declare tolower(c: bytes) -> bytes
# Convert lowercase letter to uppercase
declare toupper(c: bytes) -> bytes
```

littlewhitecloud:
Improve stdlib and fix #376
- [x] Add comment about the math lib functions
- [x] Add more math lib functions
- [ ] Write a tutorial about using functions in mem.jou (maybe not lol)
- [ ] Support more function about process.jou
- [ ] ...


littlewhitecloud:
That's strange, it will show old update history, but no change file.

littlewhitecloud:
fine, nvm.

Akuli:
These functions have surprising gotchas that users should be aware of:
- These functions use the current locale, which is os-dependent, and also some projects call `setlocale(LC_ALL, "")` or similar while others don't. The `setlocale()` call could also be hidden inside a library, e.g. GTK calls it. So with these functions, you can write code that works in project X on operating system Y, but does not work in a different project or on a different operating system. Or maybe your code breaks when you add a GUI written in GTK.
- They assume that one character fits in one byte (which in UTF-8 text is only true for ASCII characters).
- When these functions don't work as expected, you only notice it with non-ASCII strings. Programmers tend to test their code only with ASCII strings.

Many other languages (e.g. Python) instead give you information about unicode characters in a way that does not depend on the current locale. I'm not yet ready to decide what's the best approach for Jou.

littlewhitecloud:
I think jou should not depend on the current locale. (Otherwise, we will do more omplex stuff

littlewhitecloud:
> These functions have surprising gotchas that users should be aware of:
> 
> * These functions use the current locale, which is os-dependent, and also some projects call `setlocale(LC_ALL, "")` or similar while others don't. The `setlocale()` call could also be hidden inside a library, e.g. GTK calls it. So with these functions, you can write code that works in project X on operating system Y, but does not work in a different project or on a different operating system. Or maybe your code breaks when you add a GUI written in GTK.
> * They assume that one character fits in one byte (which in UTF-8 text is only true for ASCII characters).
> * When these functions don't work as expected, you only notice it with non-ASCII strings. Programmers tend to test their code only with ASCII strings.
> 
> Many other languages (e.g. Python) instead give you information about unicode characters in a way that does not depend on the current locale. I'm not yet ready to decide what's the best approach for Jou.

Think jou should not depend on setlocale.

littlewhitecloud:
List:
> - [ ] Add comment about the math lib functions
> - [ ] Write a tutorial about using functions in mem.jou (maybe not lol)
> - [ ] Support more function about process.jou
> - [ ] ...

Issue #383, opened 2023-04-08
We can't set variable value outside the function


littlewhitecloud:
I am working on the stdlib and I write code like this:
```python
# System functions
import "stdlib/str.jou"

declare getenv(varname: byte*) -> byte*

def is_windows() -> bool:
    return getenv("OS") != NULL and strstr(getenv("OS"), "Windows") != NULL

def is_macos() -> bool:
    return getenv("OSTYPE") != NULL and starts_with(getenv("OSTYPE"), "darwin")

def platform() -> byte*:
    return getenv("OS")

WINDOWS = is_windows()
MACOS = is_macos()
```
and it tell me:
```
jou sys.jou
compiler error in file "sys.jou", line 15: expected a definition or declaration, got a variable name 'WINDOWS'
The process failed with status 1.
```
So how can I give a value to variable outside the functions?

Akuli:
Jou doesn't do something magically and hide it. If you say `WINDOWS = is_windows()`, Jou will not run the `is_windows()` function when the program starts. If you want something to run when the program starts, it needs to go to the `main()` function or `_windows_startup.jou`.

But instead of having these functions in the stdlib, it would be better if you could detect the OS at compile time rather than at runtime. Maybe there would be special compile-time constants `WINDOWS` and `MACOS`. This way it would be possible to declare functions that don't exist in all systems. For example:

```
if WINDOWS:
    declare GetModuleFileNameA(hModule: void*, lpFilename: byte*, nSize: int) -> int
elif MACOS:
    declare _NSGetExecutablePath(buf: byte*, bufsize: int*) -> int
else:
    declare readlink(linkpath: byte*, result: byte*, result_size: long) -> long

def find_current_executable() -> byte*:
    if WINDOWS:
        # use GetModuleFileNameA()
    elif MACOS:
        # use _NSGetExecutablePath() 
    else:
        # use readlink()
```

Currently this is not possible, because:
- there are no compile-time `if` statements
- there is no special `WINDOWS` constant whose value is known at compile time.

As a temporary solution that made this kinda work, I added some spaghetti with a few hard-coded function names to the compiler:

https://github.com/Akuli/jou/blob/914200d20bfbcd17d660c18689ac2f910a94c2da/src/codegen.c#L189-L196

PR #384, opened 2023-04-08
More work on stdlib

### LINKED ISSUE #376: types.jou?

littlewhitecloud:
Improve stdlib and fix #376
- [x] Add comment about the math lib functions
- [x] Add more math lib functions
- [ ] Write a tutorial about using functions in mem.jou (maybe I am not good at writing tutorials lol)
- [ ] Add more test for the new functions
- [ ] Support more functions for os.jou 
- [ ] Make output can be colorful
- [ ] * Built-in functions

\* Maybe we can put some functions into _builtin.jou and import it as a special file like _windows_startup.jou when compiling


stdlib/mem.jou
```diff
@@ -1,13 +1,49 @@
 # Memory management
 
+# TODO: repalce the long with size_t
```
Akuli, line 3: This comment doesn't belong here. Currently Jou only supports 64-bit systems, so `long` and `size_t` are always of the same size.

stdlib/mem.jou
```diff
@@ -1,13 +1,49 @@
...
+# Beacause of many function prototype use size_t:
+# For example:
+#     void * memmove (void * destination, const void * source, size_t num)
+
 # Heap allocations
+
 # TODO: write a tutorial about using these and add a link
+
+# Allocates a block of bytes of memory, returning a pointer to the beginning of the block.
```
Akuli, line 12: To be honest I don't like the comments of `malloc()`, `realloc()` and `free()`. I think all Jou users either know about these functions already (because they have experience in C), or need a full tutorial (not just a few comments) for using them.

stdlib/math.jou
```diff
@@ -1,40 +1,112 @@
+# Math lib
+#TODO: add test
```
Akuli, line 2: Which functions still need tests?

stdlib/math.jou
```diff
@@ -1,40 +1,112 @@
...
+# Rounds x toward zero, returning the nearest integral value that is not larger in magnitude than x.
+declare trunc(x: double) -> double
+# Returns the integral value that is nearest to x, with halfway cases rounded away from zero.
 declare round(x: double) -> double
+declare lround(x: double) -> long
+# Rounds x to an integral value,
+declare rint(x: int) -> double
+declare lrint(x: long) -> double
+# Rounds x to an integral value
+declare nearbyint(x: double) -> double
```
Akuli, line 29: What is the difference between `round()` and `nearbyint()`? Why do we need both?

stdlib/math.jou
```diff
@@ -1,40 +1,112 @@
...
+# fmod = numer - tquot * denom
+declare fmod(numer: double, denom: double) -> double
+# Rounds x toward zero, returning the nearest integral value that is not larger in magnitude than x.
+declare trunc(x: double) -> double
+# Returns the integral value that is nearest to x, with halfway cases rounded away from zero.
 declare round(x: double) -> double
+declare lround(x: double) -> long
+# Rounds x to an integral value,
+declare rint(x: int) -> double
+declare lrint(x: long) -> double
```
Akuli, line 27: How are these different from `foo as double`? Why do we need these?

stdlib/math.jou
```diff
@@ -1,40 +1,112 @@
...
+# Returns x*y+z.
+declare fma(x: double, y: double, z: double) -> double
 
+# Rounding and remainder functions
+# Rounds x upward, returning the smallest integral value that is not less than x.
 declare ceil(x: double) -> double
+# Rounds x downward, returning the largest integral value that is not greater than x.
 declare floor(x: double) -> double
+# Returns the floating-point remainder of numer/denom (rounded towards zero):
+# fmod = numer - tquot * denom
```
Akuli, line 18: Instead of adding the `fmod` function, we should make the `%` operator work  with floats. I'm not sure if it currently works with floats.

stdlib/math.jou
```diff
@@ -1,40 +1,112 @@
...
+declare fmod(numer: double, denom: double) -> double
+# Rounds x toward zero, returning the nearest integral value that is not larger in magnitude than x.
+declare trunc(x: double) -> double
```
Akuli, line 21: I think most people would write `x as int` or `x as long` instead of `trunc(x)`. Do we really need the `trunc()` function?

stdlib/math.jou
```diff
@@ -1,40 +1,112 @@
...
 declare round(x: double) -> double
+declare lround(x: double) -> long
+# Rounds x to an integral value,
+declare rint(x: int) -> double
+declare lrint(x: long) -> double
+# Rounds x to an integral value
+declare nearbyint(x: double) -> double
+# Returns the floating-point remainder of numer/denom (rounded to nearest):
+# remainder = numer - rquot * denom
+declare remainder(numer: double, denom: double) -> double
```
Akuli, line 32: What is the difference between `remainder()` and `fmod()`? I guess we don't need either of them if we have `%` operator for floats.

stdlib/math.jou
```diff
@@ -1,40 +1,112 @@
...
+# Returns the sine of an angle of x radians.
 declare sin(x: double) -> double
+# Returns the tangent of an angle of x radians.
 declare tan(x: double) -> double
+# Returns the principal value of the arc cosine of x, expressed in radians.
+declare acos(x: double) -> double
+# Returns the principal value of the arc sine of x, expressed in radians.
+declare asin(x: double) -> double
+# Returns the principal value of the arc tangent of, expressed in radians.
+declare atan(x: double) -> double
```
Akuli, line 46: Your explanations for math functions make sense for someone who is very familiar with the math, but maybe not for a typical programmer :) We could maybe explain them just a little bit more, so that if you don't quite remember something (e.g. what is a radian or what the 'a' prefix means), that is fine.

```suggestion
# These functions use radians as the angle unit.
# One radian is the angle of a circle slice with equal radius and arc length, about 57 degrees.
# A full turn (360 degrees) is 2pi (about 6.28) radians.
declare cos(x: double) -> double
declare sin(x: double) -> double
declare tan(x: double) -> double
# The 'a' prefixed functions are inverse functions.
# For example, asin is also known as arcsin and sin^-1.
declare acos(x: double) -> double
declare asin(x: double) -> double
declare atan(x: double) -> double
```

stdlib/math.jou
```diff
@@ -1,40 +1,112 @@
...
+# Returns the principal value of the arc tangent of, expressed in radians.
+declare atan2(x: double, y: double) -> double
```
Akuli, line 48:
```suggestion
# Returns the angle of a point (x,y). Note the reversed order of the arguments.
declare atan2(y: double, x: double) -> double
```

stdlib/math.jou
```diff
@@ -1,40 +1,112 @@
...
 
-declare fmod(x: double, y: double) -> double
-declare sqrt(x: double) -> double
-declare cbrt(x: double) -> double
+# Floating-point manipulation functions
+# Returns a value with the magnitude of x and the sign of y
+declare copysign(x: double, y: double) -> double
```
Akuli, line 52: Is there a use case for this function? I don't think I have ever used it in Python or C.

stdlib/types.jou
```diff
@@ -0,0 +1,28 @@
+# Functions for checking types
```
Akuli, line 1: As discussed in your previous PR, I'm not yet sure if I want these in Jou's stdlib.

stdlib/sys.jou
```diff
@@ -0,0 +1,12 @@
+# System functions
+import "stdlib/str.jou"
+declare getenv(varname: byte*) -> byte*
+
+def is_windows() -> bool:
+    return getenv("OS") != NULL and strstr(getenv("OS"), "Windows") != NULL
```
Akuli, line 6: As I said in #383 I would prefer checking these at compile-time rather than runtime.

stdlib/process.jou
```diff
@@ -9,6 +9,12 @@ declare system(command: byte*) -> int
 # Exit the current process.
 declare exit(status: int) -> noreturn  # Clean exit. Status 0 is success, other status (usually 1) means fail.
 declare abort() -> noreturn  # Hard exit. Similar to a crash.
+declare execl(path: byte*, arg: byte*, ...) -> int # Execute the executable file with argument list args, replacing the current process.
+
+declare access(pathname: byte*, mode: int) -> int # Use the real uid/gid to test for access to a path.
```
Akuli, line 14: This function has nothing to do with processes. Maybe we should make a separate `paths.jou` which would be similar to `pathlib` and `os.path` in Python?

stdlib/process.jou
```diff
@@ -9,6 +9,12 @@ declare system(command: byte*) -> int
 # Exit the current process.
 declare exit(status: int) -> noreturn  # Clean exit. Status 0 is success, other status (usually 1) means fail.
 declare abort() -> noreturn  # Hard exit. Similar to a crash.
+declare execl(path: byte*, arg: byte*, ...) -> int # Execute the executable file with argument list args, replacing the current process.
```
Akuli, line 12: This function doesn't exist on Windows. We can add it later when it's possible to declare a function only on some platforms.

Akuli (commented):
Didn't review second half of math.jou yet

stdlib/math.jou
littlewhitecloud: We can remove it when "%" operator work with floats.

stdlib/math.jou
littlewhitecloud: like fmod(), we will delete it when "%" operator works

stdlib/math.jou
littlewhitecloud: like 
```python
printf("%d", copysign(4, -1))
```
And it will output -4.0
I will add an example for it.

stdlib/types.jou
littlewhitecloud: In my issues #376
In the self-hosted file "token.jou"
There is a TODO:
```python
# TODO: move to stdlib
declare isprint(b: int) -> int
```

littlewhitecloud: just think it later

stdlib/process.jou
littlewhitecloud: Maybe os.jou?
Jou miss a os module

stdlib/math.jou
littlewhitecloud: The new functions.

stdlib/sys.jou
littlewhitecloud: So, what should we do?

littlewhitecloud:
Think rename ```process.jou``` to ```os.jou``` is necessary.

stdlib/types.jou
Akuli: Maybe add a `?` after the TODO:

```
# TODO: move to stdlib?
```

stdlib/sys.jou
Akuli: Compile-time if statements to check the platform, in a separate PR.

stdlib/process.jou
Akuli: Jou doesn't need to mirror Python perfectly, but maybe `os.jou` would be good because Python users expect to find some things in `os`.

stdlib/types.jou
littlewhitecloud: Fine, but I think we can add it into stdlib.

stdlib/sys.jou
littlewhitecloud: Fine, is that meaning I should remove sys.jou in this or?

Akuli: Yes.

stdlib/mem.jou
littlewhitecloud: I am not good at writing tutorial(lol)

littlewhitecloud:
image
That's strange, I can't see anything about the failed test.

Akuli:
Here's what it looks like to me:

s

littlewhitecloud:
Thanks. Maybe my network has some error. (I always cannot connect github.com)

littlewhitecloud:
```
@@ -1,44 +1 @@
-1
-1145141919180
-114.514000
-5.000000
-4.000000
-3.000000
-1.140000
-3.000000
-3
-1.140000
-0.500000
-0.707107
--0.828017
-3.141593
--1.570796
--0.785398
--2.356194
--4.000000
--0.000000
-10.067662
-27.289917
-1.000000
-4.499[686](https://github.com/Akuli/jou/actions/runs/4649342960/jobs/8227618542?pr=384#step:5:687)
-4.499933
--0.549306
-20.085537
-8.000000
-40.000000
-4.736198
-2.056905
-6.832890
-5.140000
-1.140000
-0.000000
-5.814141
-0.760806
-4.000000
-2.236068
-3.000000
-5.000000
-0.842[701](https://github.com/Akuli/jou/actions/runs/4649342960/jobs/8227618542?pr=384#step:5:702)
-0.157299
-120.000000
-4.787492
+clang: error: linker command failed with exit code 1 (use -v to see invocation)
  Error: Compilers behave differently when given "tests/should_succeed/mathlibtest.jou".
```
?

Akuli:
:D

I'll try this on a linux computer at some point.

littlewhitecloud (commented):
Think ```self_hosted``` and ```tests``` doesn't have problems.

littlewhitecloud:
I'm not sure we should add color output function to ```io.jou``` or just put it into ```colorio.jou```

littlewhitecloud:
Should jou have built-in functions?

Akuli:
This PR is becoming pretty large. Can you instead make multiple smaller PRs? For example, I would like to review and merge all `math.jou` changes separately.

I don't think we should have built-in functions or color printing in stdlib, at least not yet. What problems would they solve?

Akuli:
If you make a new branch before starting a pull request, you can keep multiple pull requests up at once.

Issue #385, opened 2023-04-16
Compile time if statements


Akuli:
I want:

```
if WINDOWS:
    declare GetModuleFileNameA(hModule: void*, lpFilename: byte*, nSize: int) -> int
elif MACOS:
    declare _NSGetExecutablePath(buf: byte*, bufsize: int*) -> int
else:
    declare readlink(linkpath: byte*, result: byte*, result_size: long) -> long

def find_current_executable() -> byte*:
    if WINDOWS:
        # use GetModuleFileNameA()
    elif MACOS:
        # use _NSGetExecutablePath() 
    else:
        # use readlink()
```

There is currently no good way to write platform-specific code, and I have ended up with some amusing workarounds:

https://github.com/Akuli/jou/blob/914200d20bfbcd17d660c18689ac2f910a94c2da/src/codegen.c#L189-L196

Previously mentioned in https://github.com/Akuli/jou/issues/383#issuecomment-1500832302

PR #386, opened 2023-04-16
Refactor AST

Akuli:
The idea is to refactor things so that a file contains `AstStatement`s directly. This way I can e.g. parse `if` statements that are not inside a function, needed for #385.

### DIFF
--- a/self_hosted/ast.jou
+++ b/self_hosted/ast.jou
@@ -412,6 +412,11 @@ enum AstStatementKind:
     InPlaceMultiply # x *= y
     InPlaceDivide   # x /= y
     InPlaceModulo   # x %= y
+    Function
+    Class
+    Enum
+    GlobalVariableDeclaration
+    GlobalVariableDefinition
 
 class AstStatement:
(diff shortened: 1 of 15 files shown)


self_hosted/ast.jou
```diff
@@ -474,10 +482,32 @@ class AstStatement:
...
+            if self->function.body.nstatements == 0:
+                printf("declare a function: ")
+            else:
+                printf("define a function: ")
+            self->function.print(tp)
+        elif self->kind == AstStatementKind::Class:
+            printf("define a ")
+            self->classdef.print(tp)
+        elif self->kind == AstStatementKind::Enum:
+            printf("Define ")
```
Akuli, line 495:
```suggestion
            printf("define ")
```

self_hosted/parser.jou
```diff
@@ -1,3 +1,4 @@
+import "stdlib/io.jou"
```
Akuli, line 1:
```suggestion
```

PR #387, opened 2023-04-16
More mathlib

littlewhitecloud:
Improve & extend mathlib


littlewhitecloud:
Still failed on compare-compilers, that strange.
@Akuli could you please take a time to test it on Linux?

Akuli:
The problem was that the self-hosted compiler didn't pass `-lm` when linking, so the math functions didn't work. Apparently it works on Windows without `-lm` for whatever reason.

stdlib/math.jou
```diff
@@ -1,40 +1,107 @@
...
+declare fma(x: double, y: double, z: double) -> double
 
+# Rounding and remainder functions
+# Rounds x upward, returning the smallest integral value that is not less than x.
 declare ceil(x: double) -> double
+# Rounds x downward, returning the largest integral value that is not greater than x.
 declare floor(x: double) -> double
+# Returns the floating-point remainder of numer/denom (rounded towards zero):
+# fmod = numer - tquot * denom
+declare fmod(numer: double, denom: double) -> double
```
Akuli, line 18: The `%` operator does something similar enough already:

```suggestion
```

stdlib/math.jou
```diff
@@ -1,40 +1,107 @@
...
-declare fmod(x: double, y: double) -> double
-declare sqrt(x: double) -> double
-declare cbrt(x: double) -> double
+# Floating-point manipulation functions
+# Returns a value with the magnitude of x and the sign of y
+# Example:
+#   printf("%f", copysign(4, -1))
+# Output:
+#   -4.0
+declare copysign(x: double, y: double) -> double
```
Akuli, line 47: I still don't know when this function would be useful.

stdlib/math.jou
```diff
@@ -1,40 +1,107 @@
...
+# Returns a quiet NaN (Not-A-Number) value of type .
+# The NaN values are used to identify undefined or non-representable values for floating-point elements, such as the square root of negative numbers or the result of 0/0.
+# The argument can be used by library implementations to distinguish different NaN values in a implementation-specific manner.
```
Akuli, line 50: The description is confusing, because it doesn't explain what a "library implementation" is, what is "a implementation-specific manner", and most importantly, it doesn't tell how to use the function.

```suggestion
# Use nan("") to get a quiet NaN (Not-A-Number) value.
# The argument must be an empty string.
```

stdlib/math.jou
```diff
@@ -1,40 +1,107 @@
...
+declare nan(tagp: byte*) -> double
+# Returns the next representable value after x in the direction of y.
+declare nextafter(x: double, y: double) -> double
```
Akuli, line 53: Have you ever used this function? Can you think of an example of when it would be useful?

stdlib/math.jou
```diff
@@ -1,40 +1,107 @@
...
+# Returns the area hyperbolic tangent of x.
+declare atanh(x: double) -> double
 
+# Exponential and logarithmic functions
+# Returns the base-e exponential function of x
 declare exp(x: double) -> double
 declare exp2(x: double) -> double
-declare pow(x: double, y: double) -> double
+# Returns the result of multiplying x (the significand) by 2 raised to the power of exp (the exponent).
 declare ldexp(x: double, exp: int) -> double
```
Akuli, line 74: Similar to `fma()`, I don't think we need this.

```suggestion
```

stdlib/math.jou
```diff
@@ -1,40 +1,107 @@
...
+declare acosh(x: double) -> double
+# Returns the area hyperbolic sine of x.
+declare asinh(x: double) -> double
+# Returns the area hyperbolic tangent of x.
+declare atanh(x: double) -> double
 
+# Exponential and logarithmic functions
+# Returns the base-e exponential function of x
 declare exp(x: double) -> double
 declare exp2(x: double) -> double
```
Akuli, line 72:
```suggestion
declare exp(x: double) -> double  # returns e^x
declare exp2(x: double) -> double  # returns 2^x
```

stdlib/math.jou
```diff
@@ -1,40 +1,107 @@
...
-declare pow(x: double, y: double) -> double
+# Returns the result of multiplying x (the significand) by 2 raised to the power of exp (the exponent).
 declare ldexp(x: double, exp: int) -> double
-
+# Returns the natural logarithm of x.
+declare log(x: double) -> double
```
Akuli, line 76: This comment should only apply to the log() function:
```suggestion
declare log(x: double) -> double  # Returns the natural logarithm of x.
```

stdlib/math.jou
```diff
@@ -1,40 +1,107 @@
...
+declare log2(x: double) -> double
+# Returns e raised to the power x minus one
 declare fmax(x: double, y: double) -> double
 declare fmin(x: double, y: double) -> double
+# Returns the positive difference between x and y.
+declare fdim(x: double, y: double) -> double
+# Returns e raised to the power x minus one
+declare expm1(x: double) -> double
+# Returns the natural logarithm of one plus x.
+declare log1p(x: double) -> double
```
Akuli, line 87: Once again, I don't see how these functions are useful, because it's easy to achieve what they do with other functions.

```suggestion
```

tests/should_succeed/mathlibtest.jou
```diff
@@ -5,27 +5,57 @@ def main() -> int:
...
+    printf("%f\n", asin(-1.0)) # Output: -1.570796
+    printf("%f\n", atan(-1.0)) # Output: -0.785398
+    printf("%f\n", atan2(-1.0, -1.0)) # Output: -2.356194
+
+    printf("%f\n", copysign(4.0, -1.0)) # Output: -4.000000
+    printf("%f\n", nextafter(0.0, -1.0)) # Output: -0.000000
+
+    printf("%f\n", cosh(3.0)) # Output: 10.067662
+    printf("%f\n", sinh(4.0)) # Output: 27.289917
+    printf("%f\n", tanh(45.0)) # Output: 1.000000
```
Akuli, line 31: I don't like the 45 in these tests, because it looks confusingly like an angle, but that's not what it is. For example, the `tanh()` function calculates `(exp(x)-exp(-x)) / (exp(x)+exp(-x))`, where `exp(45)` is really large and `exp(-45)` is really small; therefore the division is close to 1. This has nothing to do with 45-degree angles.

stdlib/math.jou
```diff
@@ -1,40 +1,107 @@
...
+# Rounds x upward, returning the smallest integral value that is not less than x.
 declare ceil(x: double) -> double
+# Rounds x downward, returning the largest integral value that is not greater than x.
 declare floor(x: double) -> double
+# Returns the floating-point remainder of numer/denom (rounded towards zero):
+# fmod = numer - tquot * denom
+declare fmod(numer: double, denom: double) -> double
+# Returns the integral value that is nearest to x, with halfway cases rounded away from zero.
 declare round(x: double) -> double
+declare lround(x: double) -> long
```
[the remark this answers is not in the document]
Akuli: Ahh, it rounds. I get now why the `round()` function is useful: `round(1.7)` is `2.0` while `1.7 as int` would be `1`. But we should still delete `lround()`.

littlewhitecloud:
Change test later

stdlib/math.jou
```diff
@@ -1,40 +1,79 @@
...
 declare fabs(x: double) -> double
 
+# Rounding and remainder functions
+# Rounds x upward, returning the smallest integral value that is not less than x.
 declare ceil(x: double) -> double
+# Rounds x downward, returning the largest integral value that is not greater than x.
 declare floor(x: double) -> double
+# Returns the integral value that is nearest to x, with halfway cases rounded away from zero.
 declare round(x: double) -> double
+declare lround(x: double) -> long
```
littlewhitecloud, line 16:
```suggestion

```

stdlib/math.jou
littlewhitecloud: Useless, so I will delete it.
```suggestion

```

stdlib/math.jou
littlewhitecloud: Fine, this function is useless I will remove it...

littlewhitecloud:
So... Should ```jou``` have ```types.jou```?

Akuli:
Not yet, for now it's probably better if you `declare` whatever you need. Otherwise Jou programmers will import the `<ctype.h>` functions without really knowing what they do, and without being aware of their problems (depending on the locale and assuming that a character fits in one byte).

stdlib/math.jou
```diff
@@ -1,40 +1,73 @@
...
+# A full turn (360 degrees) is 2pi (about 6.28) radians.
+declare cos(x: double) -> double
+declare sin(x: double) -> double
+declare tan(x: double) -> double
+# The 'a' prefixed functions are inverse functions.
+# For example, asin is also known as arcsin and sin^-1.
 declare acos(x: double) -> double
 declare asin(x: double) -> double
 declare atan(x: double) -> double
+# Returns the angle of a point(x, y). Note the reversed order of the arguments.
```
Akuli, line 30: Nit:
```suggestion
# Returns the angle of a point (x, y). Note the reversed order of the arguments.
```

Akuli (approved):
Thanks! I have one more tiny suggestion.

PR #388, opened 2023-04-21
Compile time if

### LINKED ISSUE #385: Compile time if statements

Akuli:
Fixes #385

### DIFF
--- a/src/evaluate.c
+++ b/src/evaluate.c
@@ -112,11 +112,67 @@ static void handle_toplevel_statement(AstStatement *stmt, StatementList *result)
     }
 }
 
+static void simplify_if_statements_in_body(const AstBody *body);
+
+static void simplify_if_statements_in_statement(AstStatement *stmt)
+{
+    switch(stmt->kind) {
+    case AST_STMT_DEFINE_CLASS:
+        for (AstClassMember *m = stmt->data.classdef.members.ptr; m < End(stmt->data.classdef.members); m++)
+            if (m->kind == AST_CLASSMEMBER_METHOD)
+                simplify_if_statements_in_body(&m->data.method.body);
+        break;
+    case AST_STMT_FUNCTION:
+        simplify_if_statements_in_body(&stmt->data.function.body);
+        break;
+    case AST_STMT_FOR:
+        simplify_if_statements_in_body(&stmt->data.forloop.body);
+        break;
+    case AST_STMT_WHILE:
+        simplify_if_statements_in_body(&stmt->data.whileloop.body);
+        break;
+    case AST_STMT_IF:
+        simplify_if_statement(&stmt->data.ifstatement);
+        for (int i = 0; i < stmt->data.ifstatement.n_if_and_elifs; i++)
+            simplify_if_statements_in_body(&stmt->data.ifstatement.if_and_elifs[i].body);
+        simplify_if_statements_in_body(&stmt->data.ifstatement.elsebody);
+        break;
+    case AST_STMT_BREAK:
+    case AST_STMT_ASSERT:
+    case AST_STMT_ASSIGN:
+    case AST_STMT_CONTINUE:
+    case AST_STMT_DECLARE_GLOBAL_VAR:
+    case AST_STMT_DECLARE_LOCAL_VAR:
+    case AST_STMT_DEFINE_ENUM:
+    case AST_STMT_DEFINE_GLOBAL_VAR:
+    case AST_STMT_EXPRESSION_STATEMENT:
+    case AST_STMT_INPLACE_ADD:
+    case AST_STMT_INPLACE_DIV:
+    case AST_STMT_INPLACE_MOD:
+    case AST_STMT_INPLACE_MUL:
+    case AST_STMT_INPLACE_SUB:
+    case AST_STMT_PASS:
+    case AST_STMT_RETURN:
+        // these cannot contain if statements, no need to recurse here
+        break;
+    }
+}
+
+static void simplify_if_statements_in_body(const AstBody *body)
+{
+    for (int i = 0; i < body->nstatements; i++)
+        simplify_if_statements_in_statement(&body->statements[i]);
+}
+
 void evaluate_compile_time_if_statements(AstFile *file)
 {
+    // If statements at top level
     StatementList result = {0};
     for (int i = 0; i < file->body.nstatements; i++)
         handle_toplevel_statement(&file->body.statements[i], &result);
     file->body.statements = result.ptr;
     file->body.nstatements = result.len;
+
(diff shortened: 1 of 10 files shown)


src/evaluate.c
```diff
@@ -0,0 +1,174 @@
...
+    case AST_STMT_DEFINE_GLOBAL_VAR:
+    case AST_STMT_DEFINE_CLASS:
+    case AST_STMT_DEFINE_ENUM:
+        Append(result, *stmt);
+        break;
+
+    default:
+        // TODO: test this
+        // TODO: better error message?
+        fail_with_error(stmt->location, "you must put this inside a function");
```
littlewhitecloud, line 107: I think the word "must" is a bit strong, so I write this error message:
```suggestion
        fail_with_error(stmt->location, "You should put this inside the function instead of put it outside the function");
```

littlewhitecloud, line 107: I think the word "must" is a bit strong, so I write this error message:
```suggestion
        fail_with_error(stmt->location, "You should put this inside the function instead of put it outside the function");
```

littlewhitecloud (changes requested):

self_hosted/bug_workaround.jou
```diff
@@ -0,0 +1,15 @@
+# TODO: these shouldn't be needed
+
+if WINDOWS:
+    def is_windows() -> bool:
+        return True
+else:
+    def is_windows() -> bool:
+        return False
```
Akuli, line 8: I think I'll fix in this PR. But there is another problem, if you do `if WINDOWS: call_some_windows_only_func()` inside a function/method that will not work, and I will probably create issue about it and fix later

self_hosted/evaluate.jou
```diff
@@ -0,0 +1,53 @@
+# Compile-time evaluating if statements.
+
+import "./ast.jou"
+import "./bug_workaround.jou"
+import "./errors_and_warnings.jou"
+import "stdlib/str.jou"
+import "stdlib/mem.jou"
+
+
+# Return valus: 1=true, 0=false, -1=unknown
```
Akuli, line 10:
```suggestion
```

self_hosted/paths.jou
```diff
@@ -2,88 +2,69 @@ import "stdlib/mem.jou"
...
+    declare mkdir(path: byte*, mode: int) -> int  # posix
 
-declare GetModuleFileNameA(hModule: void*, lpFilename: byte*, nSize: int) -> int
-declare readlink(linkpath: byte*, result: byte*, result_size: long) -> long
-declare _mkdir(path: byte*) -> int  # windows
-declare mkdir(path: byte*, mode: int) -> int  # posix
 declare dirname(path: byte*) -> byte*
 declare stat(path: byte*, buf: byte[1000]*) -> int  # lol
 declare popen(command: byte*, type: byte*) -> FILE*
 declare pclose(stream: FILE*) -> int
```
Akuli, line 22:
```suggestion
```

src/evaluate.c
```diff
@@ -0,0 +1,79 @@
+#include "jou_compiler.h"
+
+
+const struct SpecialVar specialVars[] = {
```
Akuli, line 4: this probably bad idea, simpler to call `evaluate_condition()` whenever this would be needed

src/evaluate.c
```diff
@@ -0,0 +1,79 @@
...
+// This handles nested if statements.
+void evaluate_compile_time_if_statements(AstBody *body)
+{
+    int i = 0;
+    while (i < body->nstatements) {
+        if (body->statements[i].kind != AST_STMT_IF) {
+            i++;
+        } else {
+            replace(body, i, evaluate_compile_time_if_statement(&body->statements[i].data.ifstatement));
+        }
```
Akuli, line 77: would be easier to understand if inverted

src/evaluate.c
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+    }
+
+    fail(expr->location, "cannot evaluate condition at compile time");
+}
+
+
+// returns the statements to replace if statement with
+// TODO: don't leak memory
```
Akuli, line 36: shouldn't leak no longer

```suggestion
```

src/jou_compiler.h
```diff
@@ -353,6 +356,13 @@ struct AstFile {
     AstBody body;
 };
 
+// Terminated with {NULL,false}
+extern const struct SpecialVar { const char *name; bool value; } specialVars[];
```
Akuli, line 360: should delete

Akuli:
This cause compiler leak memory will need debug
```python
if WINDOWS:
    def foo() -> bool:
        return True
else:
    def foo() -> bool:
        return False

def main() -> int:
    return 0
```

self_hosted/bug_workaround.jou
Moosems: Why not do:
```python
def is_windows() -> bool:
    if WINDOWS:
        return True
    return False
```
?

Akuli: That doesn't work yet. The code for handling `if` statements outside functions is not the same code that handles `if` statements inside functions. Only the code for `if` statements outside functions knows about `WINDOWS` and `MACOS`.

As I wrote above (though not very clearly, it was around 4AM), I will fix this before I merge this PR.

Akuli: This now works. In fact, I deleted `is_windows()` because it was unnecessary.

src/typecheck.c
```diff
@@ -64,6 +64,8 @@ static const Type *find_any_var(const FileTypes *ft, const char *name)
                 *(*var)->usedptr = true;
             return (*var)->type;
         }
+    if (get_special_constant(name) != -1)
+        return boolType;
```
Akuli, line 68: This should probably be the first thing in this function

PR #393, opened 2023-10-22
Fix TODO: the file itself cannot be imported

littlewhitecloud:


littlewhitecloud:
```c
static bool statement_conflicts_with_an_import(const AstStatement *stmt, const ExportSymbol *import)
{
    switch(stmt->kind) {
    case AST_STMT_FUNCTION:
        return import->kind == EXPSYM_FUNCTION && !strcmp(import->name, stmt->data.function.signature.name);
    case AST_STMT_DECLARE_GLOBAL_VAR:
    case AST_STMT_DEFINE_GLOBAL_VAR:
        return import->kind == EXPSYM_GLOBAL_VAR && !strcmp(import->name, stmt->data.vardecl.name);
    case AST_STMT_DEFINE_CLASS:
        return import->kind == EXPSYM_TYPE && !strcmp(import->name, stmt->data.classdef.name);
    case AST_STMT_DEFINE_ENUM:
        return import->kind == EXPSYM_TYPE && !strcmp(import->name, stmt->data.enumdef.name);
    default:
        assert(0);  // TODO
    }
}
```
Akuli, why there is a TODO with nothing?

src/main.c
```diff
@@ -365,12 +365,13 @@ static void add_imported_symbol(struct FileState *fs, const ExportSymbol *es, As
 
 static void add_imported_symbols(struct CompileState *compst)
 {
-    // TODO: should it be possible for a file to import from itself?
-    // Should fail with error?
     for (struct FileState *to = compst->files.ptr; to < End(compst->files); to++) {
         for (AstImport *imp = to->ast.imports.ptr; imp < End(to->ast.imports); imp++) {
             struct FileState *from = find_file(compst, imp->resolved_path);
             assert(from);
+            if (from->path == to->path) {
```
Akuli, line 372: This seems wrong, because in C, `==` does not compare strings properly. It checks whether they *point to the same memory address*, so if you have the same text in two different places of computer's memory, it will not be equal.

Because the `FileState`s are stored in a list, you can just check if it's the same `FileState`:

```suggestion
            if (from == to) {
```

self_hosted/main.jou
```diff
@@ -235,6 +235,9 @@ class Compiler:
                     if strcmp(self->files[isrc].ast.path, imp->resolved_path) == 0:
                         src = &self->files[isrc]
                         break
+                if (src->ast.path == dest->ast.path):
```
Akuli, line 238: I think the same also applies in Jou. Also, you don't need parentheses in Jou if statements. (I think there's an issue about generating a warning about unnecessary parentheses.)

```suggestion
                if src == dest:
```

Akuli (commented):
Thanks for working on this!

src/main.c
littlewhitecloud: Thank you.

self_hosted/main.jou
littlewhitecloud: LOL, I write too much C

Akuli (approved):
Nice :)

PR #414, opened 2023-12-03
handle import same file twice

### LINKED ISSUE #81: Double import
This is not an error, but it should be:

```python
from "stdlib/io.jou" import printf, printf
```

Akuli:
Fixes #81

Maybe I will add error instead of warning...

### DIFF
--- a/self_hosted/main.jou
+++ b/self_hosted/main.jou
@@ -247,7 +247,7 @@ class Compiler:
                         message: byte[500]
                         snprintf(message, sizeof(message), "file \"%s\" is imported twice", imp->specified_path)
                         fail(imp->location, message)
-                seen_before[k] = src
+                seen_before[i] = src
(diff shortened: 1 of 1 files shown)


self_hosted/main.jou
```diff
@@ -235,10 +237,17 @@ class Compiler:
...
                 if src == dest:
                     fail(imp->location, "the file itself cannot be imported")
 
-                assert src != NULL
+                for k = 0; k < i; k++:
+                    if seen_before[k] == src:
+                        message: byte[500]
+                        snprintf(message, sizeof(message), "file \"%s\" is imported twice", imp->specified_path)
+                        fail(imp->location, message)
+                seen_before[k] = src
```
Akuli, line 250:
```suggestion
                seen_before[i] = src
```

PR #418, opened 2023-12-04
is_ascii_letter

littlewhitecloud:
I am lazy to write tests lol


Akuli:
`isupper()` and similar functions depend on the current locale, which is nowadays a bad idea. That's why I wanted functions similar to other stuff in `ascii.jou` instead.

littlewhitecloud:
> `isupper()` and similar functions depend on the current locale, which is nowadays a bad idea. That's why I wanted functions similar to other stuff in instead.`ascii.jou`

I am sorry I forgot that. I will rewrite it without `declare`

tests/should_succeed/ascii_test.jou
```diff
@@ -10,6 +10,10 @@ def main() -> int:
     assert is_ascii_digit('0')
     assert is_ascii_digit('7')
     assert is_ascii_digit('9')
+    assert is_ascii_letter('A')
+    assert is_ascii_letter('Z')
+    assert is_ascii_letter('a')
+    assert is_ascii_letter('z')
```
Akuli, line 16: A few nits:
- Can you also add a couple tests starting with `assert not is_ascii_letter`? For example, it is important that `\0` is not considered an ASCII letter.
- It would be nice to have first all `is_ascii_digit` tests, then all `is_ascii_letter` tests. Now `is_ascii_letter` tests are in the middle of `is_ascii_digit` tests.
- In the future, please make two separate PRs when you do two things (e.g. add `is_ascii_letter` and upgrade LLVM). You can do this by making a new branch.

littlewhitecloud: I will sep these two things, but it will cause segfault when using LLVM 15

littlewhitecloud: https://github.com/Akuli/jou/actions/runs/7101569253/job/19329964038

Akuli (approved):
Awesome :)

PR #429, opened 2023-12-06
Compile-time-if related refactoring

Akuli:
- Support `not`, `and`, `or` operators in compile-time if statements. For example, `if WINDOWS or MACOS:`
- Move platform-specific errno stuff from `_windows_startup.jou`/`_macos_startup.jou` to `errno.jou`
- Move stdin/stdout/stderr handling from `_windows_startup.jou` and `_macos_startup.jou` into a new file `_io_init.jou`
- Delete `_windows_startup.jou` and `_macos_startup.jou`


Akuli:
This breaks self-hosted compiler, but only on macos, because global variables are not implemented yet and IO on macos needs them. I think I'll merge this after I add support for global variables.

PR #435, opened 2023-12-09
.gitinore in jou_compiled

### LINKED ISSUE #397: jou_compiled gitignore
Would be nice if `jou_compiled/.gitignore` was auto-generated and it ignored everything. This is super easy to generate, too. The file should just contain a single character `*`.

Other tools with output folders do this already, e.g. `wasm-pack`.

littlewhitecloud:
fix #397

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -107,8 +107,10 @@ static void mkdir_exist_ok(const char *p)
 static void write_gitinore(const char *p)
 {
     char *filename = malloc_sprintf("%s/.gitinore", p);
-    if (access(filename, F_OK))
+    if (access(filename, F_OK)) {
+        free(filename);
         return;
+    }
     FILE *gitinore = fopen(filename, "w");
     fprintf(gitinore, "*");
 
--- a/src/output.c
+++ b/src/output.c
@@ -106,7 +106,7 @@ static void mkdir_exist_ok(const char *p)
 
 static void write_gitinore(const char *p)
 {
-    char *filename = malloc_sprintf("%s/.gitinore", p);
+    char *filename = malloc_sprintf("%s/.gitignore", p);
     if (access(filename, F_OK)) {
         free(filename);
         return;
--- a/src/output.c
+++ b/src/output.c
@@ -112,9 +112,10 @@ static void write_gitinore(const char *p)
         return;
     }
     FILE *gitinore = fopen(filename, "w");
-    fprintf(gitinore, "*");
-
-    fclose(gitinore);
+    if (gitinore != NULL) {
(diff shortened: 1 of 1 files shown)


Moosems:
@Akuli I think it looks fine, what about you?

src/output.c
```diff
@@ -103,6 +104,20 @@ static void mkdir_exist_ok(const char *p)
     exit(1);
 }
 
+static void write_gitinore(const char *p)
+{
+    char *filename = malloc_sprintf("%s/.gitinore", p);
+    if (access(filename, F_OK)) {
+        free(filename);
+        return;
+    }
```
Akuli, line 113: I think we don't need to check if file exists. It doesn't matter to write it many times.

```suggestion
```

src/output.c
```diff
@@ -103,6 +104,20 @@ static void mkdir_exist_ok(const char *p)
...
+    FILE *gitinore = fopen(filename, "w");
+    fprintf(gitinore, "*");
+
+    fclose(gitinore);
```
Akuli, line 117: On Windows, opening the file for writing fails if another process already has it opened. This can happen in various surprising ways:
- another instance of Jou compiler is currently writing `*` into it
- user's editor is running git in the background
- some other program is reading the `.gitignore` (many programs do).

If you don't check for NULL, trying to use `gitinore` will crash the compiler when opening fails.

```suggestion
    if (gitinore != NULL) {
        fprintf(gitinore, "*");
        fclose(gitinore);
    }
```

You might think that the `access()` check solves this, but it doesn't quite. Another program could create the file between `access()` and `fopen()`.

Akuli (commented):
This is a bit trickier than I thought, but not too bad.

src/output.c
Akuli: Nevermind. On Windows, it makes sense to avoid opening the file unnecessarily, because it prevents other programs from using the file.

src/output.c
```diff
@@ -103,6 +104,20 @@ static void mkdir_exist_ok(const char *p)
     exit(1);
 }
 
+static void write_gitinore(const char *p)
+{
+    char *filename = malloc_sprintf("%s/.gitinore", p);
```
Akuli, line 109:
```suggestion
    char *filename = malloc_sprintf("%s/.gitignore", p);
```

Akuli (commented):
hmmm

littlewhitecloud:
myyyyy fault

littlewhitecloud:
thanks

Akuli (approved):
Thanks!

PR #454, opened 2023-12-14
More work on self hosted compiler

Akuli:
I believe I am pretty close to making the self-hosted compiler compile itself.


Akuli:
Something weird happens with booleans on Windows if optimizations are enabled when compiling the self-hosted compiler. It ends up getting a boolean `foo` that behaves so that `foo as int` evaluates to 255, not 0 or 1 as expected. Also, `if ... or foo` does not run when this happens.

This will take a while to minimize into something small, so I can make a bug report / question to LLVM. It is difficult, because the self-hosted compiler is creating bool 255, and it is a lot of code.

Moosems:
> I believe I am pretty close to making the self-hosted compiler compile itself.

Yay!!

Akuli:
This PR became outdated and I'm working on improving self-hosted compiler differently, by copying over `.c` files and fixing them until everything works. This means I automatically get all features in self-hosted.

PR #437, opened 2023-12-10
Add more tests to fix todos

littlewhitecloud:

Issue #257, opened 2023-02-25
Passing arguments to Jou programs


Akuli:
Currently not possible, unless you first compile to a file with `jou -o filename.exe program.jou` and then pass arguments like so `./filename.exe arg1 arg2 arg3`

littlewhitecloud:
we don't need passing arguments to Jou programs when compile. There is no programming language gives args to programs when compile

Akuli:
Indeed. If you use `-o`, it causes Jou to only compile the program and not run it, so it shouldn't accept additional arguments if `-o` is used. But without `-o` it compiles and then runs the program, and that's what this issue is about.

Moosems:
> we don't need passing arguments to Jou programs when compile. There is no programming language gives args to programs when compile

And why can't Jou be the first 😄? I think it could actually have some amazing applications.

Akuli:
C compilers have the `-D` option, which basically specifies a compile-time argument. If you pass `-Dfoo=123` to the compiler, then every occurrence of the variable name `foo` in the `.c` files will be replaced by 123, as if the C code contained `#define foo 123`.

I think Jou should have something similar, but it is somewhat low priority. You could instead auto-generate a file that gets imported into Jou. In fact, I already do that:

https://github.com/Akuli/jou/blob/b709670adaad11dec65b742716f6dd0c78b57015/Makefile.posix#L33-L36

Moosems:
Is there another way besides just overriding every occurrence of a variable?

Akuli:
In C there really isn't, and with properly named variables, it isn't as bad as you would expect it to be.

For Jou, the best option would probably be reading environment variables at compile time. Environment variables are a "standard" way to pass info to programs that most developers are already familiar with (and if not, will be very useful to learn). This way compiling the program would look something like `MY_VAR=123 jou program.jou`.

Moosems:
Hmmm. The reason I ask is because I could imagine myself wanting to use compile-time optimizations for code that is dependent one one thing or another such as platform or whatnot that can be determined dynamically in the code when speed isn't a priority but can be optimized for distribution.

Akuli:
You can already do `if WINDOWS:` or `if MACOS:` at compile time. That will probably need to be generalized at some point in the future.

Moosems:
I know, I was just using that as an example.

Akuli:
I think this should be a compiler flag, like this:

```
jou filename.jou --args "arg1 arg2 arg3"
```

[...]

PR #456, opened 2023-12-16
More tests

littlewhitecloud:
Something goes wrong in #437 so I open a new pull request, sorry.
- [x] make self-hosted compiler output the same results
- [x] add more tests


littlewhitecloud:
- [ ] tokenize

src/typecheck.c
```diff
@@ -87,9 +87,8 @@ static const char *short_type_description(const Type *t)
         case TYPE_ARRAY:
             return "an array type";
         case TYPE_BOOL:
-            // TODO: Is it possible to get this in an error message?
-            return "the built-in bool type";
-    }
+            assert(0);
```
Akuli, line 90: It is actually possible to get this in an error message:
```
def foo() -> None:
    x = True.asdf()
```

littlewhitecloud: Fine, I never thought user may run the code like this.

littlewhitecloud:
Find a lot of todos in each file and add tests for them if they said they need tests

src/typecheck.c
```diff
@@ -246,8 +245,6 @@ static Signature handle_signature(FileTypes *ft, const AstSignature *astsig, con
     else
         sig.returntype = type_from_ast(ft, &astsig->returntype);
 
-    // TODO: validate main() parameters
-    // TODO: test main() taking parameters
```
Akuli, line 250: We could keep these todo comments until #257 is fixed.

littlewhitecloud: Really? The self-hosted compiler can get args now

Akuli: Good point. We should still add validation: if main has parameters, there must be 2 of them, of types `int` and `byte**`.

Akuli (approved):
Thanks!

PR #462, opened 2023-12-16
Port the class field pointer hell to self-hosted

Akuli:
Pointers in classes are always `i8*` in LLVM IR, so that a class can contain a pointer to itself. This way you can implement a linked list, for example.

Previously this wasn't done in self-hosted compiler.


Akuli:
Something weird happens with booleans on Windows if optimizations are enabled when compiling the self-hosted compiler. It ends up getting a boolean `foo` that behaves so that `foo as int` evaluates to 255, not 0 or 1 as expected. Also, `if ... or foo` does not run when this happens.

This will take a while to minimize into something small, so I can make a bug report / question to LLVM. It is difficult, because the self-hosted compiler is creating bool 255, and it is a lot of code.

PR #464, opened 2023-12-17
O1 is the default in the help message

### LINKED ISSUE #463: help message optimization default wrong
https://github.com/Akuli/jou/blob/06c74f5b1e022f5283aa0c11a879893423423e4b/src/main.c#L43

Default was changed to 1 after this was written

littlewhitecloud:
fix #463
I am sorry I forgot to change that

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -57,7 +57,7 @@ void parse_arguments(int argc, char **argv)
     memset(&command_line_args, 0, sizeof command_line_args);
     command_line_args.argv0 = argv[0];
     /* Set default optimize to O1, user sets optimize will overwrite the default flag*/
-    command_line_args.optlevel = 1; 
+    command_line_args.optlevel = 1;
 
     if (argc == 2 && !strcmp(argv[1], "--help")) {
         // Print help.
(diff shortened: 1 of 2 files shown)


src/main.c
```diff
@@ -32,6 +32,7 @@ static void optimize(LLVMModuleRef module, int level)
     LLVMDisposePassManager(pm);
 }
 
+/* Sync with tests/should_succeed/compiler_cli.jou */
```
Akuli, line 35: IMO we don't need this comment. You will notice if tests fail.

src/main.c
```diff
@@ -40,7 +41,7 @@ static const char help_fmt[] =
     "\n"
     "Options:\n"
     "  -o OUTFILE       output an executable file, don't run the code\n"
-    "  -O0/-O1/-O2/-O3  set optimization level (0 = default, 3 = runs fastest)\n"
+    "  -O0/-O1/-O2/-O3  set optimization level (1 = default, 3 = runs fastest)\n"
```
Akuli, line 44: We could explain what 0 does:

```suggestion
    "  -O0/-O1/-O2/-O3  set optimization level (0 = no optimization, 1 = default, 3 = runs fastest)\n"
```

Akuli (approved):
Thanks!

Akuli:
> I am sorry I forgot to change that

It's ok, things like this happen to everyone all the time :)

PR #476, opened 2023-12-23
Fix out of bounds memory access bug in split_by_ascii_whitespace()

Akuli:


tests/should_succeed/ascii_test.jou
```diff
@@ -53,4 +53,15 @@ def main() -> int:
...
     free(words)
 
+    s = ""
+    words = split_by_ascii_whitespace(s)
+    assert words[0] == NULL
+    free(words)
+
+    s = "test1 test2"
+    words = split_by_ascii_whitespace(s)
+    printf("%s %s\n", words[0], words[1])  # Output: test1 test2
```
Akuli, line 63:
```suggestion
    puts(words[0])  # Output: test1
    puts(words[1])  # Output: test2
```

PR #484, opened 2023-12-26
Methods that take self by value

### LINKED ISSUE #483: Methods that take self by value
Sometimes it would be nice to make a method that takes `self` not as a pointer, but as a value. For example, `BigInt` in #482 has no methods, and instead there are functions like `def bigadd(x: BigInt, y: BigInt) -> BigInt`. I did it that way because `foo.add(bar).add(baz)` didn't work (needs a pointer to `foo.add(bar)`), but `bigadd(bigadd(foo, bar), baz)` does work.

Maybe I should do something like this:

```python
class Foo:
    x: int
    y: int

    # Method that takes self a pointer
    def print1(self) -> None:
        printf("%d %d\n", self->x, self->y)

    # Same thing, different syntax
    def print1(self: Foo*) -> None:
        printf("%d %d\n", self->x, self->y)

    # Method that takes self by value
    def print1(self: Foo) -> None:
        printf("%d %d\n", self.x, self.y)
```

This would also kinda fix #439. You could use `.` if you want, but it's usually not what you want and usually you do `->` instead.

Akuli:
Fixes #483

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -155,7 +155,11 @@ static AstSignature parse_function_signature(ParserState *ps, bool accept_self)
         } else if (is_keyword(ps->tokens, "self")) {
             if (!accept_self)
                 fail(ps->tokens->location, "'self' cannot be used here");
-            AstNameTypeValue self_arg = { .name="self", .name_location=ps->tokens++->location };
+            AstNameTypeValue self_arg = { .name="self", .type.kind = AST_TYPE_NAMED, .type.data.name = "", .name_location=ps->tokens++->location };
+            if (is_operator(ps->tokens, ":")) {
+                ps->tokens++;
+                self_arg.type = parse_type(ps);
+            }
             Append(&result.args, self_arg);
             used_self = true;
         } else {
--- a/src/print.c
+++ b/src/print.c
@@ -192,7 +192,7 @@ static void print_ast_function_signature(const AstSignature *sig)
 
 static void print_ast_call(const AstCall *call, struct TreePrinter tp, const AstExpression *self);
 
-static void print_ast_expression(const AstExpression *expr, struct TreePrinter tp)
+void print_ast_expression(const AstExpression *expr, struct TreePrinter tp)
 {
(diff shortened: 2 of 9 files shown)


src/parse.c
```diff
@@ -155,7 +155,11 @@ static AstSignature parse_function_signature(ParserState *ps, bool accept_self)
         } else if (is_keyword(ps->tokens, "self")) {
             if (!accept_self)
                 fail(ps->tokens->location, "'self' cannot be used here");
-            AstNameTypeValue self_arg = { .name="self", .name_location=ps->tokens++->location };
+            AstNameTypeValue self_arg = { .name="self", .type.kind = AST_TYPE_NAMED, .type.data.name = "", .name_location=ps->tokens++->location };
+            if (is_operator(ps->tokens, ":")) {
+                ps->tokens++;
+                self_arg.type = parse_type(ps);
```
Akuli, line 161: There should be error if self type is something other than `Class` or `Class*`

src/print.c
```diff
@@ -192,7 +192,7 @@ static void print_ast_function_signature(const AstSignature *sig)
 
 static void print_ast_call(const AstCall *call, struct TreePrinter tp, const AstExpression *self);
 
-static void print_ast_expression(const AstExpression *expr, struct TreePrinter tp)
+void print_ast_expression(const AstExpression *expr, struct TreePrinter tp)
```
Akuli, line 195:
```suggestion
static void print_ast_expression(const AstExpression *expr, struct TreePrinter tp)
```

src/types.c
```diff
@@ -196,8 +196,29 @@ Type *create_enum(const char *name, int membercount, char (*membernames)[100])
...
+            case TYPE_CLASS:
+                return sig->argtypes[0];
+            default:
+                assert(0);
+        }
+    }
+    return NULL;
+}
+
+const Type *self_is_passed_as_pointer(const Signature *sig)
```
Akuli, line 211: this ever used? probably not

PR #488, opened 2023-12-26
Fix comparing two unsigned types

### LINKED ISSUE #478: byte < byte
```
import "stdlib/io.jou"

def main() -> int:
    printf("%d\n", (100 as byte) < (200 as byte))
    return 0
```

Expected output: 1

Actual output: 0

Akuli:
Fixes #478

### DIFF
--- a/tests/should_succeed/compare_bytes.jou
+++ b/tests/should_succeed/compare_bytes.jou
@@ -1,0 +1,18 @@
+import "stdlib/io.jou"
+
+
+def main() -> int:
+    a = 0 as byte
+    b = 100 as byte
+    c = 200 as byte
+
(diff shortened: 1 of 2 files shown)


tests/should_succeed/compare_bytes.jou
```diff
@@ -0,0 +1,18 @@
+import "stdlib/io.jou"
+
+
+def main() -> int:
```
Akuli, line 4:
```suggestion
# TODO: It's not possible to test things like "signed < unsigned" yet with types of the same size.
# For 8-bit there's only unsigned (byte), for 16-bit only signed (short), 32-bit only signed (int), 64-bit only signed (long).
def main() -> int:
```

PR #491, opened 2023-12-26
Support LLVM 14

Akuli:


Akuli:
LLVM 15 might need more work, so merging this as is.

Some notes:
- Apparently Windows already used clang 14, and probably LLVM 14 too.
- There are new deprecation warnings. I'll probably fix them in a new PR.

PR #495, opened 2024-02-15
Add experimental support for NetBSD

taahol:
* On NetBSD, compile against libLLVM installed as a package.
* Support diff(1) implementations lacking the option --color.
* ~~Check for core dumps in compare_compilers.sh.~~


Makefile.posix
```diff
@@ -7,14 +7,18 @@ LLVM_CONFIG ?= $(shell \
...
 CFLAGS += $(shell $(LLVM_CONFIG) --cflags)
 CFLAGS += -gdwarf-4  # https://github.com/llvm/llvm-project/issues/56550
 LDFLAGS ?= $(shell $(LLVM_CONFIG) --ldflags --libs)
 
 ifeq ($(CC),cc)
 	# default c compiler --> use clang
-	CC := $(shell $(LLVM_CONFIG) --bindir)/clang
+	CC := $(shell \
+	which `$(LLVM_CONFIG) --bindir`/clang \
+	|| which clang \
```
Akuli, line 20: This is repeated in a couple places. Maybe make a new variable for it, or just use `$(CC)` everywhere.

README.md
```diff
@@ -164,6 +164,12 @@ MacOS support is new. Please create an issue if something doesn't work.
 
 </details>
 
+<details> <summary>NetBSD</summary>
+
+Additionally, there is support for NetBSD on an experimental
+basis. Packages `clang`, `libLLVM`, `gmake` and `bash` are required.
```
Akuli, line 170: I forgot that these are inside `<details>` sections. IMO full instructions would be better (copy/paste from linux section as needed), because it's still easy to ignore the parts that you don't care about.

compare_compilers.sh
```diff
@@ -87,7 +92,7 @@ echo "Compiling the self-hosted compiler..."
 if [[ "$OS" =~ Windows ]]; then
     mingw32-make self_hosted_compiler.exe
 else
-    make self_hosted_compiler
+    $(which gmake || which make) self_hosted_compiler
```
Akuli, line 95: Maybe this would be more readable: (can't leave a github suggestion here)

```sh
elif which gmake >/dev/null; then
    # NetBSD
    gmake self_hosted_compiler
else
    make self_hosted_compiler
```

Also applies to other scripts.

doctest.sh
```diff
@@ -22,10 +22,15 @@ if [[ "$OS" =~ Windows ]]; then
     mingw32-make
     jou="$PWD/jou.exe"
 else
-    make
+    eval $(which gmake || which make)
     jou="$PWD/jou"
 fi
 
+DIFF=$(which gdiff || which diff)
+if [[ x"$($DIFF --help | grep -- --color)" != x ]]; then
```
Akuli, line 30: You could maybe use `grep -q` if it's portable enough. Not sure.

src/update.c
```diff
@@ -61,7 +61,7 @@ void update_jou_compiler()
         fail();
 #else
     confirm("Run \"git pull && make\"?");
```
Akuli, line 63: On NetBSD it would be nice if this said `Run "git pull && gmake"?`

stdlib/_netbsd_startup.jou
```diff
@@ -0,0 +1,22 @@
+# A call to the _jou_netbsd_startup() function is inserted to the
+# start of every Jou program when compiling for NetBSD.
+
+# On NetBSD, these std* things are macros
+# expanding to elements of the __sF array
```
Akuli, line 5:
```suggestion
# expanding to elements of the __sF array
# and the FILE struct is 152 bytes.
```

tests/should_succeed/compiler_cli.jou
```diff
@@ -88,7 +88,7 @@ def main() -> int:
 
     # Compile a GUI program
     if not (WINDOWS or MACOS):
-        ret = system("./jou -o /dev/null --linker-flags \"-lX11\" examples/x11_window.jou")
+        ret = system("./jou -o /dev/null --linker-flags \"-L/usr/X11R7/lib -lX11\" examples/x11_window.jou")
```
Akuli, line 91: Would be nice, if netbsd specific stuff was under `if NETBSD`. That way it's clear why the weird folder path. IMO it doesn't matter if you copy/paste rest of the command twice.

tests/should_succeed/file.jou
```diff
@@ -5,7 +5,7 @@ def write_hello_123() -> None:
     f = fopen("tmp/tests/hello.txt", "w")
     if f == NULL:
         printf("can't write tmp/tests/hello.txt\n")
-        abort()
+        exit(1) # abort() would make a core dump when comparing compilers
```
Akuli, line 8:
```suggestion
        exit(1)  # on NetBSD, abort() would make a core dump when comparing compilers
```

tests/should_succeed/compiler_cli.jou
taahol: It’s not weird, more like the Unix standard :)

tests/should_succeed/file.jou
taahol: Not specific to NetBSD but to your `ulimit` setting.

stdlib/_netbsd_startup.jou
taahol: thanks!

.github/workflows/netbsd.yml
```diff
@@ -0,0 +1,26 @@
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  test:
+    runs-on: macos-latest
```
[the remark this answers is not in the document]
taahol: > macOS runners are, in general, preferred. They support hardware accelerated nested virtualization, making them significantly faster than the Linux runners. This only applies when the runner architecture and the guest architecture are the same, in this case x86-64.

Makefile.posix
taahol: I’ll do that.

src/update.c
taahol: I don’t know, maybe?

doctest.sh
taahol: Ah, I did have a little feeling that something like that must exist. `-q` it is, indeed!

.github/workflows/netbsd.yml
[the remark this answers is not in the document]
Akuli:
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/netbsd.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: cross-platform-actions/action@v0.22.0
```
Akuli, line 12:
```suggestion
      - uses: cross-platform-actions/action@v0.23.0
```

compare_compilers.sh
```diff
@@ -5,6 +5,7 @@
 # If tokenizing/parsing a Jou file fails, both compilers should fail with the same error message.
 
 set -e
+ulimit -c unlimited
```
Akuli, line 8: For `compare_compilers` it would be better to disable core dumps entirely.

runtests.sh
```diff
@@ -206,7 +213,12 @@ function run_test()
 
     printf "\n\n\x1b[33m*** Command: %s ***\x1b[0m\n\n" "$command" > $diffpath
 
-    if diff --text -u --color=always <(
+    DIFF=$(which gdiff || which diff)
+    if $DIFF --help | grep -q -- --color; then
+        diff_color="--color=always"
+    fi
```
Akuli, line 219: Let's move this outside `run_test` function

src/main.c
```diff
@@ -461,6 +461,10 @@ int main(int argc, char **argv)
 #ifdef __APPLE__
     include_special_stdlib_file(&compst, "_macos_startup.jou");
 #endif
+#ifdef __NetBSD__
+    assert(sizeof(FILE) == 152);
```
Akuli, line 465:
```suggestion
    assert(sizeof(FILE) == 152);  // magic number in the startup file
```

PR #507, opened 2024-04-21
Use another way to get the assertion (seems bad)

littlewhitecloud:
Now using `get_assertion_from_tokens` instead of `read_assertion_from_file`.

Also need to write it on the self_hosted compiler. 

By the way, my vscode autoformatted the file, do you mind...?

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -8,7 +8,8 @@
 #include <stdio.h>
 #include <string.h>
 
-typedef struct {
+typedef struct
+{
     const Token *tokens;
     const char *stdlib_path;
     bool is_parsing_method_body;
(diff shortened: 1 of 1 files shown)


littlewhitecloud:
I'm not sure if I do something wrong with the string, hope you will check it! Thanks!

Akuli:
Thanks for working on this!

I'm not sure what the best/proper way is. With this PR, tests fail with this error:

```diff
@@ -228,7 +228,7 @@
 |     | `--- [line 87] post-increment
 |     |   `--- [line 87] get variable "s"
 |     `--- [line 87] 0 (8-bit unsigned)
-|--- [line 89] assert "result_ptr == & result [ count ]"
+|--- [line 89] assert "result_ptr == &result[count]"
 | `--- [line 89] eq
 |   |--- [line 89] get variable "result_ptr"
 |   `--- [line 89] address of
```

So your code shows `& foo [ bar ]` when the assertion contains `&foo[bar]`. We shouldn't always add spaces between tokens.

But I think there's no good way to come up with rules when to put spaces. For example, `assert a+b == c` and `assert a + b == c` are both valid, and IMO it would be good if the assertion error message showed the code exactly like it is in file, with the same spacing. This way you can use Ctrl+F or similar to find the failing assert, for example.

I need to think about this for a while. Ping me in a couple days if I haven't responded :)

littlewhitecloud:
You're right, we can't decision that users will write down what code because they can decide add spaces in the expression or not. So, maybe the way we are using now is a good way to show the code.

littlewhitecloud:
Nvm, just closing it.

PR #509, opened 2024-07-09
User-friendly error message for missing main function

### LINKED ISSUE #210: user-friendly error message for missing main func
Running `jou` on an empty file gives:

```
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/crt1.o: in function `_start':
(.text+0x20): undefined reference to `main'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

The same error happens on any other file that doesn't have a `main` function.

IMO there should be:
- An error that says clearly there is no main function, if the file you try to run doesn't define a `main` function.
- An error if you try to import a file that defines a `main` function.

littlewhitecloud:
fix #210

- [x] An error that says clearly there is no main function, if the file you try to run doesn't define a main function.
- [x] An error if you try to import a file that defines a main function.


src/codegen.c
```diff
@@ -514,8 +516,19 @@ LLVMModuleRef codegen(const CfGraphFile *cfgfile, const FileTypes *ft)
...
+    for (CfGraph **g = cfgfile->graphs.ptr; g < End(cfgfile->graphs); g++) {
+        if (strcmp((*g)->signature.name, "main") == 0)
+            mainflag = true;
         codegen_function_or_method_def(&st, *g);
+    }
+
+    if (checkmain && mainflag == false) {
+        struct Location l = {
+            .filename = cfgfile->filename,
+            .lineno = 1,
```
Akuli, line 528: If you don't want to show a line number at all, you can set `.lineno = 0` here and use an `# Output:` comment in tests.

littlewhitecloud:
```diff
-missing `main` function to execute the program
+compiler error in file "tests/other_errors/missing_main_function.jou": missing `main` function to execute the program
```
It seems that `Output` doesn't match the error.

littlewhitecloud: But the test didn't pass

littlewhitecloud:
@Akuli
This pr will make `tests\should_succeed\import_cycle.jou` fail

tests/other_errors/import_cycle.jou
```diff
@@ -6,17 +6,7 @@ def a(x: int) -> None:
...
-# Output: a 8
-# Output: b 7
-# Output: a 6
-# Output: b 5
-# Output: a 4
-# Output: b 3
-# Output: a 2
-# Output: b 1
-# Output: a 0
+# Output: compiler error in file "tests/other_errors/import_cycle.jou": imported file should not have `main` function
```
Akuli, line 9: This is valid code. It shouldn't fail to compile. It is fine to have a `main` function in an imported file if that is the file name you pass to the Jou compiler.

Maybe instead of this:

```
for each imported file:
    if file has a `main` function:
        error
```

the logic should be more like this:

```
if file is the file given to Jou compiler as argument:
    if no main function:
        error
else:
    if it has a main function:
        error
```

littlewhitecloud: > This is valid code. It shouldn't fail to compile. It is fine to have a `main` function in an imported file if that is the file you give to the Jou compiler when you compile your program.
> 
> Maybe instead of this:
> 
> ```
> for each imported file:
>     if file has a `main` function:
>         error
> ```
> 
> the logic should be more like this:
> 
> ```
> if file is the file given to Jou compiler as argument:
>     if no main function:
>         error
> else:
>     if it has a main function:
>         error
> ```

So it may cause:
```
compiler error in file "tests/other_errors/import_file_with_main.jou", line 4: a function named 'main' already exists
```

Akuli: I'm not completely sure how this should work. I'll pour more time into it during the next few days.

Akuli:
@littlewhitecloud Sorry that it took so long for me to start working on this :)

I am now quite happy with how it works. C code is reasonably clean, error messages are good, existing and new tests are passing.

But I haven't updated the self-hosted compiler. Can you do that? :)

littlewhitecloud:
> @littlewhitecloud Sorry that it took so long for me to start working on this :)
> 
> I am now quite happy with how it works. C code is reasonably clean, error messages are good, existing and new tests are passing.
> 
> But I haven't updated the self-hosted compiler. Can you do that? :)

I need some time to know what changed :) But now I'm a little busy and I will be back to school soon XD
So I am not sure if I can update the self-hosted compiler, Sorry.

littlewhitecloud:
Ok I just got time but is there `find_file` function in self-hosted compiler?

Akuli:
This loop in `main.jou` is basically the `find_file` function. It is not in a separate function, so you can either copy/paste the loop, or make a new function (or method) that does it.

```python
            found = False
            for i = 0; i < self->nfiles; i++:
                if strcmp(self->files[i].ast.path, item.path) == 0:
                    found = True
                    break
```

littlewhitecloud:
What's the return type @Akuli ?
```
def find_file(files: FileState*, nfiles: int, path: byte*) -> ?:
    for i = 0; i < nfiles; i++:
        if strcmp(files[i].ast.path, path) == 0:
            return files[i]
    return NULL
```

self_hosted/main.jou
```diff
@@ -90,6 +90,10 @@ def parse_args(argc: int, argv: byte**) -> CommandLineArgs:
 
     return result
 
+def find_file(files: FileState*, nfiles: int, path: byte*) -> FileState:
```
Akuli, line 93: You can return a pointer to an individual file, just like the C code. (See Jou tutorial or ask me to clarify if you are new to pointers.) By using pointers, you can use `NULL` as a "not found" value, and the compiler is happy because you always return something.

```suggestion
def find_file(files: FileState*, nfiles: int, path: byte*) -> FileState*:
```

self_hosted/main.jou
```diff
@@ -90,6 +90,10 @@ def parse_args(argc: int, argv: byte**) -> CommandLineArgs:
...
+    for i = 0; i < nfiles; i++:
+        if strcmp(files[i].ast.path, path) == 0:
+            return files[i]
```
Akuli, line 96: If returning a pointer:
```suggestion
            return &files[i]
    return NULL
```

self_hosted/main.jou
```diff
@@ -467,6 +483,13 @@ def main(argc: int, argv: byte**) -> int:
         compiler.process_imports_and_exports()
         compiler.typecheck_stage3_all_files()
 
+        mainfile = find_file(compiler.files, compiler.nfiles, args.main_path)
+        assert mainfile
```
Akuli, line 487: Jou is a more explicit language than C when it comes to checking for NULL. In C, `if(some_pointer)` is same as `if(some_pointer != NULL)`. In Jou you need to compare to NULL directly:

```suggestion
        assert mainfile != NULL
```

self_hosted/main.jou
```diff
@@ -467,6 +483,13 @@ def main(argc: int, argv: byte**) -> int:
...
+
+        if (not check_main_function(&mainfile.ast)):
```
Akuli, line 489: Nit: Jou does not need as much parentheses as C.
```suggestion
        if not check_main_function(&mainfile.ast):
```

Akuli:
I have added a few review comments. Let me know if you need more help :)

self_hosted/main.jou
Akuli: Here `&files[i]` is a pointer to `files[i]` (a `FileState`), not a copy of the whole `FileState`.

self_hosted/main.jou
littlewhitecloud: Thanks!

self_hosted/main.jou
littlewhitecloud: Sorry, get used to it

self_hosted/main.jou
littlewhitecloud: I forgot it add it after test

littlewhitecloud:
> I have added a few review comments. Let me know if you need more help :)

Thanks!

Akuli (approved):
Looks good, thanks!

PR #511, opened 2024-07-18
Improve Mathlib

littlewhitecloud:


stdlib/math.jou
```diff
@@ -68,14 +72,25 @@ declare asinh(x: double) -> double
...
+
+# Returns 2^x
 declare exp2(x: double) -> double
+declare exp2f(x: float) -> float
+
+# Returns 10^x
+declare exp10(x: double) -> double
+declare exp10f(x: float) -> float
+
 # Returns the natural logarithm of x.
```
Akuli, line 87: While we're here, we might as well clean up this comment. It only applies to `log()` and `logf()`. Other functions should have similar comments with the correct logarithm base, e.g. `# Returns the base 2 logarithm of x.` for `log2()` and `log2f()`.

stdlib/math.jou
```diff
@@ -39,6 +40,9 @@ declare ceil(x: double) -> double
 declare floor(x: double) -> double
 # Returns the integral value that is nearest to x, with halfway cases rounded away from zero.
 declare round(x: double) -> double
```
Akuli, line 42: You could also add float versions of `ceil`, `floor`, `round` and `fabs`. (They just get `f` to end of name, e.g. the float version of `fabs` is `fabsf`.)

Akuli:
This looks good, thanks for working on this!

I will "later" figure out why the CI is red. The changes don't look like they should break anything :)

stdlib/math.jou
```diff
@@ -68,14 +72,25 @@ declare asinh(x: double) -> double
...
 declare exp2(x: double) -> double
+declare exp2f(x: float) -> float
+
+# Returns 10^x
+declare exp10(x: double) -> double
+declare exp10f(x: float) -> float
+
 # Returns the natural logarithm of x.
 declare log(x: double) -> double
+decalre logf(x: float) -> float
```
Akuli, line 89:
```suggestion
declare logf(x: float) -> float
```

Akuli (commented):
Yeah this is why CI fails

stdlib/math.jou
littlewhitecloud: Ok, I will add the float versions of these functions.

stdlib/math.jou
littlewhitecloud: LOL, I spell `declare` keyword wrong. Thanks!

Akuli:
Thanks for working on this! I pushed a couple commits and I will merge soon.

(If you are wondering why I deleted `fmod` and `fmodf`, it is because Jou's `%` operator already does the same thing. In C, the `%` operator is only for integers, and it cannot be used with floats or doubles.)

PR #512, opened 2024-07-21
Get assertion improvement

littlewhitecloud:


littlewhitecloud:
Now the biggest headache is figuring out where to add spaces...

Akuli:
You could maybe read the file from first token's location to last token's location. When you read text from file, it already comes with spaces in the right places.

Akuli:
> maybe read the file from first token's location to last token's location

Now I remember why this is not possible. Jou's tokens do not contain column numbers, only line numbers. So you don't know where exactly the `assert` keyword ends and the assertion itself starts.

That's why the old code just reads the whole line and removes the `assert` word:

```
    if(!strncmp(str.ptr, "assert",6))
        memmove(str.ptr, &str.ptr[6], strlen(&str.ptr[6]) + 1);
```

So if the first 6 characters in the string are `"assert"`, it moves the remaining characters (including the special `\0` terminator) to start of the string. Not very nice.

Maybe some day Jou's tokens will contain a better location than just line number. Until then, I don't think this can be improved much.

littlewhitecloud:
OK, closing.

PR #513, opened 2024-07-28
Read assertion from file with offsets

Akuli:
The fix for ugly assertion reading. Adds file offsets to tokens, so that `fseek()` can be used to go to where the assertion starts and ends in the file. This way we can read the file contents after tokenizing.

- [ ] Fix weird error that only happens on Windows
- [ ] Port to self-hosted

### DIFF
--- a/src/jou_compiler.h
+++ b/src/jou_compiler.h
@@ -91,7 +91,6 @@ struct Token {
         TOKEN_OPERATOR,
         TOKEN_END_OF_FILE,  // Marks the end of an array of Token
     } type;
-    Location location;
     union {
         int16_t short_value; // TOKEN_SHORT
         int32_t int_value;  // TOKEN_INT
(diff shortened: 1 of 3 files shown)


Akuli:
This doesn't work, because of the `unread_byte()` function and Windows CRLF (`\r\n`) line endings.

When tokenizing, it is possible to unread a byte that was previously read from the file. For example, to read the `printf` from `printf("hi")`, you read bytes one at a time until you get a byte that is not valid in a name, in this case `(`. Then you unread the `(` so that the next read will return it.

The problem is unreading `"\r\n"` (Windows newline), which is translated to `'\n'` when reading, so that `read_byte()` consumes the `\r` and the `\n`, then returns just the `\n`. If we unread a `\n`, then we don't know whether the file contained `\n` or `\r`.

I could seek the file to check if it contains a `\r` before the `\n`, but then you can't compile unseekable files (such as output of another command), which would be annoying.

A better fix would be to read the entire source code into a string before tokenizing it.

PR #521, opened 2024-12-02
Add quicksort example

Akuli:


examples/quicksort.jou
```diff
@@ -0,0 +1,133 @@
...
+    #           [9]     >=pivot
+    arr[length / 2] = arr[length - 1]
+    length--
+
+    # Rearrange elements so that:
+    #   - arr[0..end_of_small] <= pivot
+    #   - arr[start_of_big..pivot_temp] >= pivot
+    #   - the above two ranges cover the entire array (they may overlap)
+    #
+    # This looks like:
```
Akuli, line 67: Picture is better than words.

```suggestion
    # Rearrange elements like this:
```

PR #531, opened 2024-12-12
Document my AI+Jou experiment

Akuli:
Tl;dr: I wrote some Jou code with an LLM that didn't know about Jou, asked what it thinks about Jou, and got a really nice response.


Moosems:
I will note that it doesn't think classes exist in the language 😄

Issue #546, opened 2025-01-05
NetBSD tests failure


Akuli:
When I make a pull request, NetBSD tests fail. See for example https://github.com/Akuli/jou/actions/runs/12615061996/job/35154688113

<details><summary>Copy/pasted output</summary>

```
Run cross-platform-actions/action@v0.23.0
Setting up VM
Running command
/usr/bin/ssh -t runner@cross_platform_actions_host sh -c 'cd "/home/runner/work/jou/jou" && exec "$SHELL" -e'
Pseudo-terminal will not be allocated because stdin is not a terminal.
download error: http://cdn.NetBSD.org/pub/pkgsrc/packages/NetBSD/amd64/9.3/All/llvm-18.1.8nb3.tgz truncated
xmlcatmgr-2.2nb1: copying /usr/pkg/share/examples/xmlcatmgr/catalog.etc.sgml to /usr/pkg/etc/sgml/catalog
xmlcatmgr-2.2nb1: copying /usr/pkg/share/examples/xmlcatmgr/catalog.etc.xml to /usr/pkg/etc/xml/catalog
xmlcatmgr-2.2nb1: copying /usr/pkg/share/examples/xmlcatmgr/catalog.share.sgml to /usr/pkg/share/sgml/catalog
xmlcatmgr-2.2nb1: copying /usr/pkg/share/examples/xmlcatmgr/catalog.share.xml to /usr/pkg/share/xml/catalog
downloading pkg_summary.bz2 done.
calculating dependencies...done.

6 packages to refresh:
  xmlcatmgr-2.2nb1 zstd-1.5.6 readline-8.2nb2 libuuid-2.32.1nb2 libffi-3.4.6
  lz4-1.10.0

1 package to upgrade:
  libxml2-2.12.9nb3

8 packages to install:
  clang-18.1.8nb3 libLLVM-13.0.1nb4 gmake-4.4.1 diffutils-3.10
  python312-3.12.8nb1 perl-5.40.0nb1 llvm-18.1.8nb3 sqlite3-3.47.2

6 to refresh, 1 to upgrade, 8 to install
[20](https://github.com/Akuli/jou/actions/runs/12615061996/job/35154688113#step:3:21)9M to download, 1307M to install

downloading lz4-1.10.0.tgz done.
downloading xmlcatmgr-2.2nb1.tgz done.
downloading zstd-1.5.6.tgz done.
downloading sqlite3-3.47.2.tgz done.
[...]

PR #549, opened 2025-01-05
Update netbsd.yml

taahol:
* Bump NetBSD version to 10.0
* Bump action version to 0.26.0

Closes: #546


Akuli:
Thanks for fixing this so quickly! :)

PR #558, opened 2025-01-07
Fix valgrind

### LINKED ISSUE #557: Running tests with valgrind failed on Tue Jan 07 2025
See valgrind output here: https://github.com/Akuli/jou/actions/runs/12645033422

Akuli:
- Last night, valgrind ran out of memory. Do not enable memory limit when running valgrind, because it consumes more memory than the program by itself, and seems to be somewhat unpredictable.
- If one valgrind run fails, do not immediately stop the others. It can be useful to know which runs fail, e.g. if only one of them fails, it is probably random/undeterministic and doesn't fail every time.

Fixes #557


Akuli:
I ran the valgrind workflow manually on the `valgrind-fix` branch and it is green.

Issue #116, opened 2023-01-21
self-hosted compiler?!!?!??


Akuli:
i.e. a Jou compiler that compiles Jou code, and of course can compile itself.

progress:

- [x] Hello world tokenizes
- [x] All test files tokenize
- [x] Hello world parses
- [x] All test files parse
- [x] Hello world typechecks
- [x] All test files typecheck
- [x] Hello world compiles to an object file
- [x] All test files compile to object files
- [x] Hello world runs
- [x] All test files run correctly
- [ ] Self-hosted compiler emits the correct control flow analysis warnings/errors (e.g. unreachable code, trying to use an undefined variable, missing return statement)
- [x] Self-hosted compiler can compile itself (?????!!!!!)
- [x] Self-hosted compiler is the compiler that ships in Windows zip files
- [x] On Linux, the `jou` executable that `make` produces is the self-hosted compiler

For now, the main way to develop the self-hosted compiler is running the `./compare_compilers.sh` script (previously named `tokenizers.sh`). It attempts to compile files with both compilers and compares the results. There are lists of various "known not working" files in `self_hosted/`, and `./compare_compilers.sh --fix` updates them automatically.

Akuli:
I have now written a tokenizer in Jou that is capable of tokenizing the hello world program. Here's how to run it (with the latest jou repository cloned, `self_hosted` isn't included in the Windows zip files that `jou --update` uses):

```
jou -o asd.exe -O1 self_hosted/tokenizer.jou
./asd.exe examples/hello.jou 
```

Output:

```
===== Tokens for file "examples/hello.jou" =====

Line 1:
  keyword "from"
  string "stdlib/io.jou"
  keyword "import"
  name "puts"
  newline token (next line has 0 spaces of indentation)

Line 3:
  keyword "def"
  name "main"
  operator '('
  operator ')'
  operator '->'
  keyword "int"
  operator ':'
  newline token (next line has 4 spaces of indentation)

Line 4:
  indent (+4 spaces)

Line 5:
  name "puts"
  operator '('
  string "Hello World"
  operator ')'
  newline token (next line has 4 spaces of indentation)

Line 6:
  keyword "return"
  integer 0
  newline token (next line has 0 spaces of indentation)

Line 7:
  dedent (-4 spaces)
  end of file
```

You can also get the **exactly same** output with the original tokenizer written in C:

```
jou --tokenize-only examples/hello.jou
```

[...]

PR #562, opened 2025-01-08
Port the whole Jou compiler to Jou

Akuli:
I ported the whole Jou compiler to Jou. The self-hosted compiler can now compile anything that the original Jou compiler can.

The self-hosted compiler can compile itself. In the below command, I called the result `compiler2`.

```
$ ./self_hosted_compiler -o compiler2 --linker-flags "-L/usr/lib/llvm-14/lib -lLLVM-14" self_hosted/main.jou 
$ ./compiler2 examples/hello.jou 
Hello World
```

Of course, the self-hosted compiler compiled with the self-hosted compiler can also compile the self-hosted compiler. I called it `compiler3`.

```
$ ./compiler2 -o compiler3 --linker-flags "-L/usr/lib/llvm-14/lib -lLLVM-14" self_hosted/main.jou 
$ ./compiler3 examples/hello.jou 
Hello World
```

**This is something I have wanted to achieve for quite some time.**

Caveats:
- Code is messy. I ported most of it by taking C code and modifying it until it works.
- It leaks memory. Some memory freeing code is commented out.
- It lacks the `simplify_cfg` compilation step. Apparently, the only real consequence of this is that you won't get a compiler warning if you forget a `return` statement in a function.

See also other checkboxes in #116.


Akuli:
The diff of this pull request looks big because I moved the old self-hosted compiler to new folder `self_hosted_old`.

That said, the new self-hosted compiler is quite different from the old one, so in that sense I did a lot of work:

screenshot-1736359789

PR #563, opened 2025-01-08
Add timeouts to all GitHub Actions workflows

Akuli:
This is a best practice that I forgot to do. I noticed this when I somehow managed to freeze the netbsd build in #562.

### DIFF
--- a/.github/workflows/netbsd.yml
+++ b/.github/workflows/netbsd.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


.github/workflows/netbsd.yml
```diff
@@ -7,6 +7,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    timeout-minutes: 5
```
Akuli, line 10: Netbsd tests do a lot...

```suggestion
    timeout-minutes: 10
```

Issue #566, opened 2025-01-08
Undefined value warnings/errors


Akuli:
The following program shouldn't compile:

```python
def main() -> int:
    pass
```

The compiler written in C correctly refuses to compile this:

```
compiler error in file "asd.jou", line 1: function 'main' must return a value, because it is defined with '-> int'
```

But the self-hosted compiler compiles this with no warnings, and produces and executable that invokes UB.

In general, the self-hosted compiler has no warnings for using undefined values (return value, variable, anything else). For example, the following program prints an undefined value. The compiler written in C shows a warning about it, but the self-hosted compiler Jou does not.

```python
import "stdlib/io.jou"

def main() -> int:
    x: int
    printf("%d\n", x)
    return 0
```

To fix this properly, the compiler should know whether a value is properly defined, undefined, or maybe undefined depending on e.g. user input. And to do that, we need to analyze Control Flow Graphs (CFGs).

The problem is that the self-hosted compiler doesn't have the `simplify_cfg` compilation step of the compiler written in C. I omitted `simplify_cfg` in #562.

PR #573, opened 2025-01-08
Make self-hosted compiler the default and introduce bootstrapping

Akuli:
When you type `make`, you now get:
- stage 1 compiler: the Jou compiler written in C (now known as bootstrap compiler)
- stage 2 compiler: the Jou compiler written in Jou, compiled with the Jou compiler written in C
- stage 3 compiler: the Jou compiler written in Jou, compiled with the Jou compiler written in Jou

~~Stage 1 and 2 compilers are compiled into a subfolder `bootstrap/` to keep the top level directory clean. The logic for finding Jou stdlib was adjusted to take that into account.~~

If you edit `compiler/foo.jou` and run `make` again, only the stage 3 compiler will be rebuilt. If you edit `bootstrap_compiler/foo.c` and run `make`, all stages will be rebuilt.

Succeeding tests are ran with all 3 compilers. Tests related to errors and warnings are only ran with the stage 3 compiler, because that's what users will use. You can pass `--stage1` or `--stage2` to `runtests.sh` to select the stage to compile and test (default is stage 3 and there's no flag for that).

Currently not all tests pass with the self-hosted compiler due to #566. I moved them to a new folder `broken_tests`.

I also updated documentation, scripts and GitHub workflows.

I deleted `compare_compilers.sh`.


Akuli:
I've messed around with Windows file paths for 2 hours, and I realized that if you run `folder/file.exe` on Windows, it really wants to set the working directory to `folder` in the newly created process.

I guess I won't place the bootstrapping executables into `bootstrap` subfolder like I had originally planned.

Akuli:
Finally GREEN!!!!!!!!

Issue #565, opened 2025-01-08
Clean up code of the self-hosted compiler


Akuli:
Code is very messy in some places, because in #562 I took C code and just changed the syntax to Jou. This means that the resulting code doesn't use nice C features that don't exist in Jou (e.g. `List(T)`/`Append()`), and also doesn't use nice Jou features that don't exist in C (e.g. methods).

Akuli:
All done!

PR #575, opened 2025-01-09
Refactor compiler/main.jou

Akuli:
This is important, because anyone trying to figure out how the compiler works will probably start from `main.jou` (or documentation).

Part of #565.


Akuli:
I still want to tweak this a little bit, at least split up `ast_to_object_file()`.

PR #578, opened 2025-01-10
Clean up compiler/ast.jou

Akuli:
Part of #565

### DIFF
--- a/compiler/ast.jou
+++ b/compiler/ast.jou
@@ -1,3 +1,9 @@
+# Defines data structures for Abstract Syntax Tree.
+#
+# Many class in this file have .print(), which can be used during debugging. If
+# .print() exists, you probably don't want to use .print_with_tree_printer(),
+# which is basically an internal detail of the .print() implementation.
+
 import "stdlib/io.jou"
 import "stdlib/str.jou"
 import "stdlib/mem.jou"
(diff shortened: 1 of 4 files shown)


compiler/ast.jou
```diff
@@ -1,3 +1,10 @@
+# This file defines data structures of the Abstract Syntax Tree. They are
+# constructed in parser.jou.
+#
+# Many class in this file have .print(), which can be used during debugging. If
```
Akuli, line 4:
```suggestion
# Many classes in this file have .print(), which can be used during debugging. If
```

compiler/ast.jou
```diff
@@ -1,3 +1,10 @@
...
+# .print() exists, you probably don't want to use .print_with_tree_printer(),
```
Akuli, line 5:
```suggestion
# .print() exists, you probably don't need to call .print_with_tree_printer(),
```

compiler/ast.jou
```diff
@@ -31,19 +40,15 @@ class AstType:
...
 
-    # TODO: use this
     def is_void(self) -> bool:
         return self->kind == AstTypeKind::Named and strcmp(self->name, "void") == 0
 
-    # TODO: use this
     def is_none(self) -> bool:
         return self->kind == AstTypeKind::Named and strcmp(self->name, "None") == 0
 
-    # TODO: use this
```
Akuli, line 42: Should include the code that switches to these in this PR

PR #582, opened 2025-01-10
compiler: Refactor control flow graph data structures

Akuli:
Part of #565


compiler/cf_graph.jou
```diff
@@ -0,0 +1,330 @@
...
+            free_constant(&self->constant)
+        if self->kind == CfInstructionKind::StringArray:
+            free(self->strarray.str)
+        if self->kind == CfInstructionKind::Call:
+            free_signature(&self->signature)
+        free(self->operands)
+
+    # operands should be NULL-terminated array, or NULL for empty
+    # TODO: does it ever need to be NULL?
+    # TODO: do we need this method at all?
```
Akuli, line 207: Might as well check in this PR whether this is needed.

compiler/structs.jou
```diff
@@ -87,12 +86,29 @@ class GlobalVariable:
...
 
+    def print(self) -> int:
+        if self->name[0] != '\0':
+            printf("%s", self->name)
+        else:
+            # Anonymous temporary variable created by compiler.
+            printf("$%d", self->id)
+
+    # Return how many characters print() outputs.
+    def print_width(self) -> int:
```
Akuli, line 104: Instead of returning width, would be better to just print the variable to the correct width here, and `printf()` makes that sorta thing quite easy.

compiler/build_cf_graph.jou
```diff
@@ -554,25 +552,25 @@ def build_function_or_method_call(
 
     assert sig != NULL
 
-    args: LocalVariable** = calloc(call->nargs + 2, sizeof(args[0]))
-    k = 0
+    args: LocalVariable** = malloc(sizeof(args[0]) * (call->nargs + 2))  # +1 for self, another because why not
```
Akuli, line 555: Might as well add args directly to the instruction instead of building a temporary array.

compiler/build_cf_graph.jou
```diff
@@ -912,12 +910,11 @@ def build_assert(st: State*, assert_location: Location, assertion: AstAssertion*
     argtypes[1] = get_pointer_type(byteType)
     argtypes[2] = intType
 
-    args = [
-        add_local_var(st, argtypes[0]),
-        add_local_var(st, argtypes[1]),
-        add_local_var(st, argtypes[2]),
-        NULL,
-    ]
+    args: LocalVariable** = malloc(sizeof(args[0]) * 3)
```
Akuli, line 913: Can use `add_operand()` instead of this

PR #591, opened 2025-01-10
Delete compiler/output.jou and move its contents to other files

Akuli:
Part of #565


compiler/main.jou
```diff
@@ -350,6 +350,34 @@ def optimize(module: LLVMModule*, level: int) -> None:
     LLVMDisposePassManager(pm)
 
 
+def compile_to_object_file(module: LLVMModule*) -> byte*:
```
Akuli, line 353: Should be renamed

PR #592, opened 2025-01-10
Move content of compiler/structs.jou into other files

Akuli:
Part of #565

### DIFF
--- a/compiler/command_line_args.jou
+++ b/compiler/command_line_args.jou
@@ -1,0 +1,122 @@
+import "stdlib/io.jou"
+import "stdlib/str.jou"
+import "stdlib/mem.jou"
+import "stdlib/process.jou"
+
+import "./update.jou"
+
+
+class CommandLineArgs:
+    argv0: byte*  # Program name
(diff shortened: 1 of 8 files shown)


compiler/command_line_args.jou
```diff
@@ -0,0 +1,122 @@
...
+    tokenize_only: bool  # If true, tokenize the file passed on command line and don't actually compile anything
+    parse_only: bool  # If true, parse the file passed on command line and don't actually compile anything
+    optlevel: int  # Optimization level (0 don't optimize, 3 optimize a lot)
+    infile: byte*  # The "main" Jou file (can import other files)
+    outfile: byte*  # If not NULL, where to output executable
+    linker_flags: byte*  # String that is appended to linking command
+
+# Command-line arguments are a global variable because I like it.
+# This variable is also accessed from other files because I like it.
+global command_line_args: CommandLineArgs
```
Akuli, line 22: I'm not sure if I like the global variable anymore.

Akuli: Yes, I still like it.

PR #593, opened 2025-01-10
Misc refactoring to self-hosted compiler

Akuli:
Small tweaks, didn't want to make separate PR for each.

Part of #565

### DIFF
--- a/compiler/types.jou
+++ b/compiler/types.jou
@@ -33,7 +33,7 @@ enum TypeKind:
     VoidPointer
     Array
     Class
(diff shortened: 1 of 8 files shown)


compiler/types.jou
```diff
@@ -33,7 +33,7 @@ enum TypeKind:
     VoidPointer
     Array
     Class
-    OpaqueClass  # class with unknown members. TODO when used?
+    OpaqueClass  # class with unknown members, used temporarily during compiling
```
Akuli, line 36:
```suggestion
    OpaqueClass  # class with unknown members, used temporarily during type checking
```

PR #605, opened 2025-01-12
Add markdown link checker

### LINKED ISSUE #602: Markdown links
It's easy to end up with broken links in markdown, see e.g. commit af87a01

To help with this, I created [check-markdown-links.py](https://github.com/Akuli/porcupine/blob/b6ea2acb9b4786f226dbaec6d234130ffd5788af/scripts/check-markdown-links.py), a simple Python script that validates all links in markdown files. I use it in Porcupine and I think in some other repo too, but I can't remember which one.

We should add it to Jou too. It needs to be edited a little bit for Jou, but probably not much. (The script mentions "porcupine" a couple times.)

Running the script in GitHub actions could be super simple. Something like this in `linux.yml`:

```yml
- uses: actions/setup-python@v5
  with:
    python-version: '3.13' 
- run: pip install requests
- run: python my_script.py
```

This would be pretty easy to contribute. If nobody does this, I'll do it myself eventually :)

littlewhitecloud:
fix #602 (I hope my modification works, I am not good at adding test in the .yml files)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,12 +70,12 @@ I will try my best to **keep Jou simple**,
 and not turn it into yet another big language that doesn't feel like C,
 such as C++, Zig, Rust, and many others.
 For example, the recommended way to print things will be C's `printf()` function,
-as explained in [the Jou tutorial](./doc/tutorial.md#cs-standard-library-libc).
+as explained in [the Jou tutorial](./doc/tutorial.md#c-s-standard-library-libc).
 This also means that I reject many feature requests.
 
 Jou is not intended to be memory safe, because it would make Jou more difficult to use.
 See [Jou's UB documentation](./doc/ub.md) for more discussion,
-including [thoughts on Rust](./doc/ub.md#rusts-approach-to-ub).
+including [thoughts on Rust](./doc/ub.md#rust-s-approach-to-ub).
 
 
 ## Setup
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ This also means that I reject many feature requests.
 
 Jou is not intended to be memory safe, because it would make Jou more difficult to use.
 See [Jou's UB documentation](./doc/ub.md) for more discussion,
-including [thoughts on Rust](./doc/ub.md#rust-s-approach-to-ub).
+including [thoughts on Rust](./doc/ub.md#rusts-approach-to-ub).
 
 
 ## Setup
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ I will try my best to **keep Jou simple**,
 and not turn it into yet another big language that doesn't feel like C,
 such as C++, Zig, Rust, and many others.
 For example, the recommended way to print things will be C's `printf()` function,
-as explained in [the Jou tutorial](./doc/tutorial.md#c-s-standard-library-libc).
+as explained in [the Jou tutorial](./doc/tutorial.md#cs-standard-library-libc).
 This also means that I reject many feature requests.
 
 Jou is not intended to be memory safe, because it would make Jou more difficult to use.
--- a/check-markdown-links.py
+++ b/check-markdown-links.py
@@ -1,0 +1,148 @@
+"""Check that links in markdown files point to reasonable places."""
+# Get from https://github.com/Akuli/porcupine/
+
+import argparse
+import os
+import re
+import subprocess
+import sys
+from functools import cache
+from http.client import responses as status_code_names
+from pathlib import Path
+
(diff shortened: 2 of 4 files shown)


littlewhitecloud:
https://github.com/Akuli/jou/pull/605/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L243
https://en.wikipedia.org/wiki/Bootstrapping_(compilers)
@Akuli this link seems broken

README.md
```diff
@@ -70,12 +70,12 @@ I will try my best to **keep Jou simple**,
...
 such as C++, Zig, Rust, and many others.
 For example, the recommended way to print things will be C's `printf()` function,
-as explained in [the Jou tutorial](./doc/tutorial.md#cs-standard-library-libc).
+as explained in [the Jou tutorial](./doc/tutorial.md#c-s-standard-library-libc).
 This also means that I reject many feature requests.
 
 Jou is not intended to be memory safe, because it would make Jou more difficult to use.
 See [Jou's UB documentation](./doc/ub.md) for more discussion,
-including [thoughts on Rust](./doc/ub.md#rusts-approach-to-ub).
+including [thoughts on Rust](./doc/ub.md#rust-s-approach-to-ub).
```
Akuli, line 78: The original `rusts` link is correct. If you open https://github.com/Akuli/jou/blob/main/doc/ub.md#rusts-approach-to-ub it takes you to the right place.

In the script, try changing `title.lower()` to `title.lower().replace("'", "")`.

check-markdown-links.py
```diff
@@ -0,0 +1,148 @@
...
+def find_links(markdown_file_path):
+    content = markdown_file_path.read_text(encoding="utf-8")
+
+    # TODO: do we have changelog.md in jou?
+    if markdown_file_path.name == "CHANGELOG.md":
+        # Ignore changelogs of old versions. Editing them doesn't make sense.
+        header_matches = list(re.finditer("^## ", content, flags=re.MULTILINE))
+        end_of_current_version = header_matches[1].start()
+        content = content[:end_of_current_version]
+
```
Akuli, line 34: No, we don't :)

```suggestion
```

check-markdown-links.py
```diff
@@ -0,0 +1,148 @@
...
+    for lineno, line in enumerate(content.splitlines(), start=1):
+        link_regexes = [
+            # [text](target)
+            # The text can contain spaces, and it can even be split across many lines.
+            # This also detects images ![text](target) and that's a good thing.
+            r"\]\((\S+?)\)",
```
Akuli, line 40: The bootstrap link is correctly in README, but the script doesn't see the last `)` in the link. I spent a while playing around with regexes, and this seems to work:

```python
>>> import re
>>> re.findall(r"\]\(([^\s()]+?(?:\([^\s()]+\))?)\)", "blah [test](foo) (blah [test2](bar)) [test3](bar_(baz))")
['foo', 'bar', 'bar_(baz)']
```

```suggestion
            r"\]\(([^\s()]+(?:\([^\s()]+\))?)\)",
```

Here's how the new regex works:

```
\]\(                               The ]( at start of link
    (                              Begins capture group for link target
     [^\s()]                       Target starts with something not whitespace or ( or )...
            +                      ...repeated 1 or more times non-greedy
             (?:                   Begins non-capturing group
                \(                 The ( in the wikipedia link
                  [^\s()]+         Text between ( and ) in the wikipedia link
                          \)       The ) in the wikipedia link
                            )      Ends non-capturing group
                             ?     The non-capturing group is optional (parentheses not required)
                              )    End capturing group
                               \)  The ) at end of link
```

Akuli (commented):
Thanks for working on this!

README.md
```diff
@@ -70,12 +70,12 @@ I will try my best to **keep Jou simple**,
 and not turn it into yet another big language that doesn't feel like C,
 such as C++, Zig, Rust, and many others.
 For example, the recommended way to print things will be C's `printf()` function,
-as explained in [the Jou tutorial](./doc/tutorial.md#cs-standard-library-libc).
+as explained in [the Jou tutorial](./doc/tutorial.md#c-s-standard-library-libc).
```
littlewhitecloud, line 73:
```suggestion
as explained in [the Jou tutorial](./doc/tutorial.md#cs-standard-library-libc).
```

littlewhitecloud (commented):
fixed now

Akuli:
This looks good. Thanks!

PR #606, opened 2025-01-12
Fix memory leaks

### LINKED ISSUE #577: Running tests with valgrind failed
Valgrind output is here: https://github.com/Akuli/jou/actions/runs/12683481551

Akuli:
Fixes #577


Akuli:
I ran the valgrind workflow manually. It passed.

PR #610, opened 2025-01-13
doc/classes.md improvements

### LINKED ISSUE #608: Classes documentation: explain instantiating better
In Jou, an instance of a class is nothing more than a (properly aligned) chunk of memory of the right size. This means that the `ClassName{field=value}` syntax is not the only way to instantiate classes. This should be emphasized and clarified in the docs.

Akuli:
- Explain instantiation better
- Create separate section about pointers and added more reasons why/when to use them
- Explain `sizeof`

Fixes #608

### DIFF
--- a/doc/classes.md
+++ b/doc/classes.md
@@ -218,7 +218,7 @@ Let's look at this syntax in more detail.
 
 The curly braces are used to distinguish instantiating syntax from function calls.
 This makes the Jou compiler simpler, but if you don't like this syntax,
-feel free to [create an issue](https://github,com/Akuli/jou/issues/new) to discuss it.
+feel free to [create an issue](https://github.com/Akuli/jou/issues/new) to discuss it.
 
 If you omit some class fields, they will be initialized to zero.
 Specifically, the memory used for the fields will be all zero bytes.
--- a/doc/classes.md
+++ b/doc/classes.md
@@ -67,7 +67,7 @@ def main() -> int:
 ```
 
 Alternatively, we could create the instance in one line with `p = Point{x = 12, y = 34}`.
-This syntax is explained in detail [below](#instantiating).
+This syntax is explained in detail [below](#instantiating-syntax).
 
 
 ## Pointers
--- a/doc/classes.md
+++ b/doc/classes.md
@@ -35,8 +35,18 @@ In reality, instances may be bigger than expected due to
 [padding](https://stackoverflow.com/questions/4306186/structure-padding-and-packing),
 but this can be almost always ignored.
 
-You can use e.g. `Point{x=12, y=34}` to instantiate the class,
-and the usual `.` syntax to access its fields:
+To create an instance of `Point`,
+we simply need enough memory to hold the two `int`s,
+and we need to tell the compiler to treat it as a `Point` instance.
+In other words, we simply create a variable whose type is `Point`:
+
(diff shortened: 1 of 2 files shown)


doc/classes.md
```diff
@@ -35,8 +35,18 @@ In reality, instances may be bigger than expected due to
 [padding](https://stackoverflow.com/questions/4306186/structure-padding-and-packing),
 but this can be almost always ignored.
 
-You can use e.g. `Point{x=12, y=34}` to instantiate the class,
-and the usual `.` syntax to access its fields:
+To create an instance of `Point`,
+we simply need enough memory to hold the two `int`s,
+and we need to tell the compiler to treat it as a `Point` instance.
+In other words, we simply create a variable whose type is `Point`:
```
Akuli, line 41:
```suggestion
we simply need to take enough memory to hold the two `int`s,
and we need to tell the compiler to treat that memory as a `Point` instance.
In other words, we create a variable whose type is `Point`:
```

doc/classes.md
```diff
@@ -46,30 +56,26 @@ class Point:
     y: int
 
 def main() -> int:
-    p = Point{x=12, y=34}
+    p: Point
+    p.x = 12
+    p.y = 34
     printf("%d, %d\n", p.x, p.y)  # Output: 12, 34
     p.y++
```
Akuli, line 63:
```suggestion

    p.y++
```

doc/classes.md
```diff
@@ -181,8 +211,13 @@ so `self` is not a pointer.
 
 ## Instantiating
```
Akuli, line 212:
```suggestion
## Instantiating Syntax
```

doc/classes.md
```diff
@@ -46,30 +56,26 @@ class Point:
...
+## Pointers
 
-This means that if you pass an instance of a class to a function, you get a copy,
-as if you had just passed the two integers:
+Instances of classes are often passed around as pointers.
+To understand why, let's try to make a function
+that increments the `x` coordinate of a `Point`:
 
-```python
+ ```python
```
Akuli, line 78:
```suggestion
```python
```

doc/classes.md
```diff
@@ -46,30 +56,26 @@ class Point:
...
-    printf("%d, %d\n", x, y)  # Output: 12, 34
-    return 0
-```
+## Pointers
 
-This means that if you pass an instance of a class to a function, you get a copy,
-as if you had just passed the two integers:
+Instances of classes are often passed around as pointers.
+To understand why, let's try to make a function
+that increments the `x` coordinate of a `Point`:
```
Akuli, line 76:
```suggestion
that increments the `y` coordinate of a `Point`:
```

doc/classes.md
```diff
@@ -86,6 +92,27 @@ def main() -> int:
     return 0
 ```
 
+The problem is that when we do `increment_y()`,
+we simply pass the 64 (or more) bytes of the struct to the `increment_y()` method.
```
Akuli, line 96:
```suggestion
The problem is that when we do `increment_y(p)`,
we simply pass the 64 (or more) bytes of the instance `p` to the `increment_y()` function.
```

Akuli:
The code in this PR is probably fine now, but GitHub is having issues.

PR #614, opened 2025-01-14
Drop support for LLVM versions older than 14

### LINKED ISSUE #473: compiler segfault when struct contains array of pointers to itself
This crashes the jou compiler:

```python
class Module:
    name: byte[10]
    receives_broadcasts: bool

    sources: Module*[10]
    nsources: int
    destinations: Module*[10]
    ndestinations: int

def main() -> int:
    m = Module{}
    return 0
```

Akuli:
Fixes #473


Akuli:
Reopening, because turns out that both changes are needed. This doesn't work on LLVM 11 and 13 (I tried in #615).

PR #615, opened 2025-01-14
Use LLVM i8* for all pointers

### LINKED ISSUE #473: compiler segfault when struct contains array of pointers to itself

Akuli:
Starting with LLVM 15, the type of the pointed object is fully ignored, so always set it to `i8`. LLVM documentation about this: https://llvm.org/docs/OpaquePointers.html#version-support

Fixes #473

### DIFF
--- a/bootstrap_compiler/codegen.c
+++ b/bootstrap_compiler/codegen.c
@@ -55,16 +55,15 @@ static LLVMTypeRef codegen_type(const Type *type)
     case TYPE_ARRAY:
         return LLVMArrayType(codegen_type(type->data.array.membertype), type->data.array.len);
     case TYPE_POINTER:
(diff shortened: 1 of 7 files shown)


Akuli:
For some reason, I get a broken stage 3 compiler with this PR on LLVM 11 and 13, but 14+ works fine.

Apparently old LLVMs rely on the pointer type being exactly correct somehow. This might have something to do with the weird behavior I saw in #462.

PR #616, opened 2025-01-14
Add support for LLVM 16

Akuli:
No code changes, just adding the number 16 everywhere.

I won't be upgrading to newer LLVM versions than 16 for now. For some reason, I can't install LLVM 17 or 18 from `apt`. I can install LLVM 19, and it doesn't compile because apparently `PassManagerBuilder.h` has been removed.


Akuli:
CI failure is unrelated, and caused by links in markdown not being online right now for whatever reason. Maybe I should somehow make the link checker not complain if an existing link is down, but it hasn't been down for very long.

Issue #84, opened 2023-01-19
A way to mark symbols as private


Akuli:
Currently all functions and structs are made available for other files to import. There should be a way to make functions and structs be visible only within the current file.

Maybe `static`?

```python
static def internal_util_function() -> int:
    return 123

static struct PrivateStruct:
    x: int
    y: int

def public_function() -> int:
    return internal_util_function() * 2

struct PublicStruct:
    message: byte*
    line: int
```

**Edit 18/1/2025** I decided to go with `@public` decorator. For now, it will only be used with functions, but eventually it should be used with everything that can be public:
- [x] declaring functions
- [x] defining functions
- [x] declaring global variables
- [x] defining global variables
- [x] defining classes
- [x] defining enums

I also need to add unused warnings as I implement this, similarly to #636:

- [x] warning if private function declared but not used
- [x] warning if private function defined but not used
- [x] warning if private global variable declared but not used
- [x] warning if private global variable defined but not used
- [x] warning if private class not used
- [ ] warning if method in private class not used
- [x] warning if private enum not used

littlewhitecloud:
I think on the contrary, you can use the "extern" keyword to decide which functions need to be exported.

littlewhitecloud:
I think on the contrary, you can use the "extern" keyword to decide which functions need to be exported.
Then other variables, functions, structs of the current file will not be accessed.

littlewhitecloud:
```python 
def internal_util_function() -> int:
    return 123

static struct PrivateStruct:
    x: int
    y: int

extern def public_function() -> int:
    return internal_util_function() * 2

extern struct PublicStruct:
    message: byte*
    line: int
```

littlewhitecloud:
And, the "static" keyword can use by global variable

littlewhitecloud:
Just suggestions

Akuli:
> I think on the contrary, you can use the "extern" keyword to decide which functions need to be exported. Then other variables, functions, structs of the current file will not be accessed.And then, they will be private

So basically, everything is private by default, and you can use a keyword like `extern` to make public things. This seems like a good idea.

I think I will probably name the keyword `public`, for two reasons:
- With `extern` many people would think: "what does the `extern` keyword do again? Ah now I remember, it makes the thing public."
- The `extern` keyword has a different meaning in C. Jou's `declare` and C's `extern` both correspond to the same idea: they tell the linker that something exists without defining it.

Akuli:
[...]

PR #638, opened 2025-01-17
Add `@public` decorator

Akuli:
Anything not decorated with `@public` is now private, and not visible in other files. Breaks backwards-compatibility for any project that contains code meant to be imported :)

Example:

```python
def only_for_this_file() -> None:
    ...

@public
def visible_in_files_that_import_this_file() -> None:
    ...

# main() doesn't need @public
def main() -> None:
    ...
```

Also, you can have multiple non-`@public` functions with the **same** name in different files! So you don't need to name most functions in a way that avoids conflicts, only public functions.

For now, this will only apply to functions, but I will later extend it to all other importable things too.

Part of #84


Moosems:
I think I mentioned this but I'm not a huge fan of language functions being implemented in decorators. I expect decorators to be something I can read the source of and understand given a few minutes, but something to dig into the compiler for :/

Akuli:
I know your opinion on this, but there are practical reasons to use decorator syntax:
- **Extensible:** There will be other things that can be marked similarly. Think `@inline` or `@windows_dllexport`.
- **Familiar:** People already have an idea of what decorators are: something you ignore when you read the code for the first time.
- **Supported by tools:** Python syntax highlighting works nicely with decorators, not so well with made-up syntax (e.g. Jou enums).

PR #647, opened 2025-01-19
Valgrind github config improvements

Akuli:
- Build a list of failed valgrind runs instead of showing only the first one.
- Refactor the code to use bash functions, so that it's easier to test it locally. You still need to substitute various `${{ github.foo }}` values with find/replace when testing.

### DIFF
--- a/.github/workflows/valgrind.yml
+++ b/.github/workflows/valgrind.yml
@@ -61,14 +61,46 @@ jobs:
       issues: write
     steps:
       - run: |
+          # Usage: find_issue <title>
+          # Echos issue number, if any
+          function find_issue() {
+              curl -s "https://api.github.com/repos/${{ github.repository }}/issues?state=open" \
+                  | jq -r --arg t "$1" '.[] | select(.title==$t) | .number'
+          }
+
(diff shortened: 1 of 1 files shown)


.github/workflows/valgrind.yml
```diff
@@ -61,14 +61,46 @@ jobs:
...
+          # Usage: echo body | new_issue <title>
+          function new_issue() {
+              curl -s -X POST \
+                  -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+                  -H "Accept: application/vnd.github.v3+json" \
+                  "https://api.github.com/repos/${{ github.repository }}/issues" \
+                  -d "$(jq -n --arg t "$1" --arg b "$(cat)" '{title: $t, body: $b}')"
+          }
+
+          link="- [$(date +%Y-%m-%d)](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})"
```
Akuli, line 94:
```suggestion
          link="- [$(date)](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})"
```

PR #660, opened 2025-01-20
For loops: make all parts optional

Akuli:
Now all three parts of the `for a; b; c:` are optional. Here are some examples.

```
# initialization is optional
i = 0
for ; i < 10; i++:
    printf("%d\n", i)

# condition is optional (defaults to True)
for i = 0;; i++:
    if i == 10:
        break
    printf("%d\n", i)

# increment is optional
for i = 0; i < 10;:
    printf("%d\n", i++)

# all fields are optional
i = 0
for ;;:
    if i == 10:
        break
    printf("%d\n", i++)
```

### DIFF
--- a/bootstrap_compiler/build_llvm_ir.c
+++ b/bootstrap_compiler/build_llvm_ir.c
@@ -914,7 +914,10 @@ static void build_loop(
 
     // Evaluate condition and then jump to loop body or skip to after loop.
     LLVMPositionBuilderAtEnd(st->builder, condblock);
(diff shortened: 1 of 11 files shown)


Akuli:
Before I merge this, I want to know what @littlewhitecloud, @Moosems and @taahol think.

Moosems:
I think it looks great

littlewhitecloud:
Looks good

littlewhitecloud:
Looks cool

Akuli:
I think I already know mostly what taahol thinks about this. We discussed it irl yesterday.

PR #662, opened 2025-01-21
compiler code cleanup: Attach type information to AST

Akuli:
Previously, the types were passed around separately. Type checking returned a `FileTypes`, which contained all type information about a Jou file. Now the AST contains the types, and during parsing, the type information is e.g. set to NULL.

This is the same idea as #600, but does it to all type information instead of just expressions.

~~Not ready to merge yet, because I still need to implement the same changes in the bootstrap compiler.~~


bootstrap_compiler/typecheck.c
```diff
@@ -1472,10 +1472,8 @@ static void typecheck_function_or_method_body(FileTypes *ft, const Signature *si
     ft->current_fom_types = End(ft->fomtypes) - 1;
     ft->current_fom_types->signature = copy_signature(sig);
 
-    for (int i = 0; i < sig->nargs; i++) {
-        LocalVariable *v = add_variable(ft, sig->argtypes[i], sig->argnames[i]);
-        v->is_argument = true;
-    }
+    for (int i = 0; i < sig->nargs; i++)
+        add_variable(ft, sig->argtypes[i], sig->argnames[i]);
```
Akuli, line 1476: This should probably be extracted into separate PR.

compiler/main.jou
```diff
@@ -116,31 +114,11 @@ class FileState:
...
-                }
+                self->ast.types.add_function(&es->funcsignature, &imp->used)
             case ExportSymbolKind.Type:
-                self->types.types = realloc(self->types.types, sizeof(self->types.types[0]) * (self->types.ntypes + 1))
-                assert self->types.types != NULL
-                self->types.types[self->types.ntypes++] = TypeAndUsedPtr{
-                    type = es->type,
-                    usedptr = &imp->used,
-                }
+                self->ast.types.add_type(es->type, &imp->used)
```
Akuli, line 119: Maybe do these in separate PR?

compiler/typecheck/step2_populate_types.jou
```diff
@@ -51,7 +38,10 @@ def handle_signature(ft: FileTypes*, astsig: AstSignature*, self_class: Type*) -
 
     msg: byte[500]
 
-    if ft->find_function_or_method(self_class, astsig->name) != NULL:
```
Akuli, line 54: Deleting `find_function_or_method()` would also be good in separate PR

PR #668, opened 2025-01-21
Bootstrap Jou from Git history

### LINKED ISSUE #661: Jou to C conversion script?
Maybe I should write a Python script that converts Jou code to C code? This way I could delete the bootstrap compiler written in C.

The script should probably produce (even better, print) a single C file. The Jou compiler is only about 10000 lines of Jou code, so even if the script produces a single C file, a C compiler is able to compile it no problem. Other projects that use generated C code sometimes run into limitations in C compilers, e.g. the C compiler uses too much memory, but this probably wouldn't happen in Jou.

Then the question is, is it easier to maintain a pile of C code that compiles Jou, or a Python script that converts Jou to C?

Akuli:
This means that we no longer need a compiler written in C.

Closes #661

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: macos-latest
     timeout-minutes: 10
     strategy:
-      - uses: actions/checkout@v4  # TODO: upgrade to v4 in other places
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0  # Fetch the whole Git history
       - run: ./bootstrap.sh
@@ -20,6 +20,7 @@ jobs:
           path: jou_bootstrap.exe
 
   test:
+    needs: bootstrap
     runs-on: macos-latest
     timeout-minutes: 10
     strategy:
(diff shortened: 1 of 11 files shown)


.github/workflows/macos.yml
```diff
@@ -5,7 +5,22 @@ on:
...
 jobs:
+  bootstrap:
+    runs-on: macos-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Fetch the whole Git history
+      - run: ./bootstrap.sh
+        shell: bash
```
Akuli, line 16: This is probably the default
```suggestion
```

.github/workflows/valgrind.yml
```diff
@@ -24,39 +63,29 @@ jobs:
...
 
   # Based on: https://github.com/python/typeshed/blob/9f28171658b9ca6c32a7cb93fbb99fc92b17858b/.github/workflows/daily.yml
   create-issue-on-failure:
     name: Create an issue if valgrind failed
     runs-on: ubuntu-latest
     timeout-minutes: 5
-    needs: [valgrind]
-    if: ${{ github.repository == 'Akuli/jou' && always() && needs.valgrind.result == 'failure' }}
+    needs: [bootstrap, valgrind]
+    if: ${{ github.repository == 'Akuli/jou' && always() && (needs.bootstrap.result == 'failure' || needs.valgrind.result == 'failure') }}
```
Akuli, line 88:
```suggestion
    needs: [check-if-needed, bootstrap, valgrind]
    if: ${{ github.repository == 'Akuli/jou' && always() && needs.check-if-needed.outputs.needed == 'true' && (needs.bootstrap.result == 'failure' || needs.valgrind.result == 'failure') }}
```

bootstrap.sh
```diff
@@ -0,0 +1,59 @@
...
+# Jou is always used to compile the next.
+#
+# We don't take all commits from Git history, because compiling all of them
+# would be slow. Instead, this script contains a hard-coded list. You need to
+# add a commit to the end of the list before you can use a new language feature
+# in the Jou compiler or standard library used by the compiler.
+
+set -e -o pipefail
+
+commits=(
```
Akuli, line 17:
```suggestion
# Add the latest commit on main branch to the end of the list if this script
# produces a compiler that is too old.
commits=(
```

dummy
Akuli, line 1: should delete

PR #673, opened 2025-01-21
Add unused function warning

### LINKED ISSUE #636: Unused function warning
Once #84 is implemented for functions, the compiler should warn if a non-public function is not used within the file and hence can be removed.

Example:

```python
import "stdlib/io.jou"

def foo() -> None:  # Warning: function 'foo' is not used
    printf("Foo\n")

def main() -> int:
    printf("main runs\n")  # Output: main runs
    return 0
```

For public functions, this is more difficult because a public Jou function may be called in a surprising way, e.g. from a different programming language. If I implement unused function detection for them, that will be done later, in another issue/PR, and there must be a way to disable it.

Akuli:
Fixes #636


Akuli:
I like how this PR implements a feature, but it actually removes more code than it adds.

PR #680, opened 2025-01-21
GitHub Actions: include .a files in Windows bootstrap artifact

### LINKED ISSUE #679: GitHub Actions: cache lib files created by windows_setup.sh?
https://github.com/Akuli/jou/blob/9b249b5584f7188ee73d25fcdb0453c41db2c81a/windows_setup.sh#L76-L86

This is the slowest part of `windows_setup.sh`. In CI, this is done during bootstrapping, and then repeated every time the bootstrap compiler is used.

Maybe CI should upload the lib files with the bootstrap compiler, so that they don't need to be recreated many times.

Akuli:
Fixes #679

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -90,12 +90,12 @@ jobs:
     timeout-minutes: 5
     steps:
       - uses: actions/checkout@v4
-      # TODO: figure out why --small doesn't work here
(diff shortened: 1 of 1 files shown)


Akuli:
This seems to make `windows_setup.sh` about 27 seconds faster. Not a game changer, but also too bad for such a small change :)

Issue #683, opened 2025-01-24
Initializing array of array strings


Akuli:
```python
argnames: byte[100][3] = ["assertion", "path", "lineno"]
```

Expected result: no error

Actual result: `compiler error in file "...", line 475: initial value for variable of type byte[100][3] cannot be of type byte*[3]
`

Then I tried

```python
argnames: byte[100][3] = ["assertion" as byte[100], "path" as byte[100], "lineno" as byte[100]]
```

which doesn't work either: `cannot cast from type byte* to byte[100]`

This work fine but not very convenient

```python
        argnames: byte[100][3]
        strcpy(argnames[0], "assertion")
        strcpy(argnames[1], "path")
        strcpy(argnames[2], "lineno")
```

Akuli:
The proper fix would be bidirectional type inference (type context), as I wrote in #359. That feels a bit too complicated for Jou, and any other solution feels like an ugly hack that won't work in corner cases (e.g. nested arrays).

I will need to think about this a bit.

PR #686, opened 2025-01-24
Compiler refactor: replace Control Flow Graphs with builders

Akuli:
A builder defines how to do somewhat elementary things, e.g. a function/method call, conditional branch, or an equals comparison. Then a new file `ast_to_builder.jou` recurses over the AST and calls methods of the builder accordingly.

For now, there is only one builder, which builds LLVM IR. The idea is to later build a special, intentionally very limited "Jou IR" to solve #566.

Also, I really hated how the old CFG's worked. They were very similar to LLVM IR, but much shittier.

TODO:
- [x] get CI green
- [x] document this
- [x] resolve `# TODO: rename to AstToBuilder`
- [x] resolve `# TODO: add error message for this, or dynamic size?`
- [x] resolve `# TODO: https://github.com/Akuli/jou/issues/683`
- [ ] resolve `# TODO: tell llvm, if we know a function is noreturn ?`
- [ ] resolve `# TODO: useful, but doesn't belong here`
- [x] resolve `# TODO: Everything named legacy_llvm14_cast can be removed once we drop LLVM 14 support`
- [ ] resolve `# TODO: which casts are necessary on LLVM 14 and which are not?`
- [ ] resolve `# TODO: method to find field`
- [x] resolve `# TODO: do we need is_main_file?`
- [x] resolve `# TODO: delete id`
- [x] resolve `# TODO: delete double pointer`


Akuli:
Not gonna resolve all TODOs. This PR already deletes more TODOs than it adds.

Akuli:
Don't know why the tests are failing now. Maybe I'll fix finish this PR some other day :)

Akuli:
I set up an awful Windows VM to debug this. (It is awful for various reasons.) This seems like a stack overflow bug.

I am adding debug prints to `examples/aoc2024/day22/part2.jou`, and running it with `../../../jou.exe -O0 part2.jou`. It stops mid-way along a long loop, exits with status 1, and prints no error. Like this:

screenshot-1737848850

But if I add a dummy variable `foo: byte[10000]` on the stack, it stops earlier:

screenshot-1737848939

I think the problem is that the new builder is allocating stack space in the loop, instead of doing it after the loop is done. I'm glad the test setup is good enough to catch this :)

Akuli:
Apparently using `-o` and then double-clicking the executable gives a better error message on Windows.

screenshot-1737849566

screenshot-1737849728

screenshot-1737849754

Due to my terrible VM setup, I cannot copy/paste or copy files from the VM, so anyone else reading this will be stuck with Finnish screenshots. Oops :)

Akuli:
It is a stack overflow https://stackoverflow.com/a/76892354

Akuli:
It worked!!!! :)

PR #696, opened 2025-01-26
Add "match ... with ..." syntax

### LINKED ISSUE #617: Match statement and strings
I am working on adding match statements to Jou (#555). For now it will be mostly for enums. This issue is about extending it to string.

To check whether strings are equal in Jou, you call the `strcmp()` function, which loops through them byte by byte, and returns `0` if the strings are equal. For example:

```python
s = "Hello"

if strcmp(s, "Hello") == 0:
    printf("Hello there! :)\n")
elif strcmp(s, "Hi") == 0:
    printf("Hiii\n")
else:
    printf("yo\n")
```

It is tempting to hard-code the `strcmp()` function into the match statement implementation, so that you could simply write this:

```python
s = "Hello"

match s:
    case "Hello":
        printf("Hello there! :)\n")
    case "Hi":
        printf("Hiii\n")
    case _:
        printf("yo\n")
```

This would call the `strcmp()` function once or twice depending on `s`. However, to me this feels too magical for Jou. I would prefer function calls to be explicit.

Maybe you could say `match s with strcmp` to specify a function for comparing the objects:

```python
s = "Hello"

match s with strcmp:
    case "Hello":
        printf("Hello there! :)\n")
    case "Hi":
        printf("Hiii\n")
    case _:
        printf("yo\n")
```

@Moosems and @littlewhitecloud, what do you think of this?

Akuli:
Fixes #617

Example:
```python
match "hello" with strcmp:
    case "hello":
        printf("yass\n")
```

For now, the match function must return `int`, so you can't match with e.g. `starts_with()`. This might actually be a good thing, so that usage of `match` statement remains quite simple and it's not tempting to make things more complicated than needed.

### DIFF
--- a/tests/should_succeed/match.jou
+++ b/tests/should_succeed/match.jou
@@ -129,6 +129,13 @@ def main() -> int:
             printf("Hiii\n")
         case "Hello":
             printf("Hello there!\n")
+        # no "case _", that's fine
+
+    match s with strcmp:
+        case "Hi":
+            printf("Hiii\n")
+        case "Hello":
+            printf("Hello there!\n")
         case _:
             printf("something else\n")  # Output: something else
 
--- a/tests/should_succeed/match.jou
+++ b/tests/should_succeed/match.jou
@@ -1,4 +1,6 @@
 import "stdlib/io.jou"
+import "stdlib/mem.jou"
+import "stdlib/str.jou"
 
 enum Foo:
     Bar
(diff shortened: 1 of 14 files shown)


tests/should_succeed/match.jou
```diff
@@ -110,4 +112,25 @@ def main() -> int:
...
+            printf("something else\n")
+
+    strcat(s, "lol")
+    match s with strcmp:
+        case "Hi":
+            printf("Hiii\n")
+        case "Hello":
+            printf("Hello there!\n")
+        case _:
+            printf("something else\n")  # Output: something else
```
Akuli, line 133: Should also test the same without `case _`

PR #699, opened 2025-01-26
Use "match ... with ..." statements in compiler and examples

Akuli:
The code is slightly longer, but much more readable.

Currently `match some_byte:` doesn't work, but `match some_byte with bytecmp` does, where `bytecmp` is the following function:

```python
# TODO: should not be needed
def bytecmp(a: byte, b: byte) -> int:
    return (a as int) - (b as int)
```

### DIFF
--- a/compiler/tokenizer.jou
+++ b/compiler/tokenizer.jou
@@ -290,7 +290,9 @@ class Tokenizer:
         if c == '\\':
             after_backslash = self->read_byte()
             match after_backslash with bytecmp:
-                case '\0' | '\n':
+                case '\0':
+                    fail(self->location, "missing ' to end the byte literal")
+                case '\n':
                     self->location.lineno--
                     fail(self->location, "missing ' to end the byte literal")
                 case 'n':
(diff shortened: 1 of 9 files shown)


compiler/tokenizer.jou
```diff
@@ -282,115 +289,124 @@ class Tokenizer:
...
-            elif after_backslash < 0x80 and isprint(after_backslash) != 0:
-                message: byte* = malloc(100)
-                sprintf(message, "unknown escape: '\\%c'", after_backslash)
-                fail(self->location, message)
-            else:
-                fail(self->location, "unknown '\\' escape")
+            match after_backslash with bytecmp:
+                case '\0' | '\n':
+                    self->location.lineno--
+                    fail(self->location, "missing ' to end the byte literal")
```
Akuli, line 295: Lets try this. It makes more sense to me.

```suggestion
                case '\0':
                    fail(self->location, "missing ' to end the byte literal")
                case '\n':
                    self->location.lineno--
                    fail(self->location, "missing ' to end the byte literal")
```

Issue #703, opened 2025-01-26
Platform if statements inside function


Akuli:
https://github.com/Akuli/jou/blob/dd896f416f0f49a9fedf35a0f7216647b5d88060/stdlib/errno.jou#L3-L26

If I write `if WINDOWS:` inside the function and compile the code on Linux, the compiler does not ignore all the code under `if WINDOWS:` like it should. That happens only if I put `if WINDOWS:` outside any functions, at the top level.

Akuli:
Related: #537

A proper fix is to use `AstStatement` and `AstBody` for everything, so that function bodies, method bodies etc all use the same data structure. This way we can eliminate if statements inside them similarly.

Issue #497, opened 2024-02-19
Platform if statements inside class


Akuli:
I want to put `if WINDOWS` inside class to create windows specific method, useful when method only needed on one platform.

```python
import "stdlib/io.jou"

class Foo:
    def bar(self) -> None:
        printf("bar\n")
        if WINDOWS:
            self->do_windows_thing()

    if WINDOWS:  # <---- Error :(
        def do_windows_thing(self) -> None:
            printf("lol\n")


def main() -> int:
    f = Foo{}
    f.bar()
    return 0
```

PR #708, opened 2025-01-27
Use AstStatement and AstBody everywhere

### LINKED ISSUE #537: Compiler crash in `statement_conflicts_with_an_import()`
The following code (attempted AoC solution) crashes the Jou compiler.

```python
import "stdlib/ascii.jou"
import "stdlib/str.jou"
import "stdlib/io.jou"
import "stdlib/mem.jou"


class CacheEntry:
    towel: byte[30]
    value: int


avail_cache: CacheEntry[1000]
avail_cache_len: int


def count_availability(towel: byte*, patterns: byte**) -> int:
    cacheptr = NULL
    if strlen(towel) < sizeof(avail_cache[0].towel):
        # Find value from cache if exists
        for i = 0; i < avail_cache_len; i++:
            if strcmp(avail_cache[i].towel, towel) == 0:
                return cacheptr->value

        # Create new cache entry if not full
        if cacheptr == NULL and avail_cache_len < sizeof(avail_cache)/sizeof(avail_cache[0]):
            cacheptr = &avail_cache[avail_cache_len++]
            strcpy(cacheptr.towel, towel)

    if towel[0] == '\0':
        return 1

    result = 0
    for i = 0; patterns[i] != NULL; i++:
        if starts_with(towel, patterns[i]):
            result += count_availability(&towel[strlen(patterns[i])], patterns)

    if cacheptr != NULL:
        cacheptr->value = result

    return result


def main() -> int:
    avail_cache_len = 0

    f = fopen("sampleinput.txt", "r")
    assert f != NULL

    patterns_buf: byte[4096]
    ret = fgets(patterns_buf, sizeof(patterns_buf) as int, f)
    assert ret != NULL
    patterns = split_by_ascii_whitespace(patterns_buf)

    # remove trailing commas
    for i = 0; patterns[i] != NULL; i++:
        patterns[i][strcspn(patterns[i], ",")] = '\0'

    line: byte[1000]
    result = 0

    while fgets(line, sizeof(line) as int, f) != NULL:
        trim_ascii_whitespace(line)
        if line[0] == '\0':
            # skip blank line
            continue

        result += count_availability(line, patterns)

    printf("%d\n", result)  # Output: 16

    fclose(f)
    free(patterns)
    return 0
```

Compiler crash is:

```
jou: src/main.c:351: _Bool statement_conflicts_with_an_import(const AstStatement *, const ExportSymbol *): Assertion `0' failed.
```

Akuli:
The first step of solving #703 and #497. Fixes #537.


compiler/parser.jou
```diff
@@ -96,63 +96,64 @@ def determine_the_kind_of_a_statement_that_starts_with_an_expression(
...
+#                snprintf(
+#                    message, sizeof message,
+#                    "class %s already has %s named '%s'",
+#                    classdef->name, p1->kind, p1->name,
+#                )
+#                fail(p2->name_location, message)
+#
+#    free(infos)
+#
+#
```
Akuli, line 156: Done now
```suggestion
```

compiler/parser.jou
```diff
@@ -1138,6 +1259,17 @@ class Parser:
             if t->kind == TokenKind.Decorator:
                 decorate(&result, t->short_string, t->location)
 
+        # Inside a class, variable declarations become fields.
```
Akuli, line 1262: Instead of this, it would be better to set it directly to the correct type, similar to FunctionDef/MethodDef

compiler/parser.jou
```diff
@@ -212,11 +213,115 @@ def decorate(stmt: AstStatement*, decor: byte*, location: Location) -> None:
...
 
+enum WhatAreWeParsing:
+    TopLevel  # must be first, so that this is the default (zero)
+    ClassBody
+    FunctionBody
+    MethodBody
+
+
+# Errors if e.g. trying to define a class inside a function (other way around it would become a method)
+def check_statement_not_misplaced(
```
Akuli, line 224: In many cases it is possible to emit error earlier, before this function can be called. For example, when we start parsing a class, we could check if it's in a function or method or other class. This is better, because if class contains errors, we tell user not to put class inside class or whatever, so they stop trying to do it early.

compiler/typecheck/step3_function_and_method_bodies.jou
```diff
@@ -1338,10 +1338,14 @@ def typecheck_statement(state: State*, stmt: AstStatement*) -> None:
...
+            | AstStatementKind.MethodDef
             | AstStatementKind.Class
+            | AstStatementKind.ClassField
+            | AstStatementKind.ClassUnion
             | AstStatementKind.Enum
             | AstStatementKind.GlobalVariableDeclare
             | AstStatementKind.GlobalVariableDef
+            | AstStatementKind.Import
         ):
             # TODO: what happens if you try to create these? are error messages good?
```
Akuli, line 1350: Yes they are
```suggestion
```

compiler/parser.jou
```diff
@@ -212,11 +155,115 @@ def decorate(stmt: AstStatement*, decor: byte*, location: Location) -> None:
...
+            | AstStatementKind.Assign
+            | AstStatementKind.InPlaceAdd
+            | AstStatementKind.InPlaceSub
+            | AstStatementKind.InPlaceMul
+            | AstStatementKind.InPlaceDiv
+            | AstStatementKind.InPlaceMod
+        ):
+            match what_are_we_parsing:
+                case WhatAreWeParsing.TopLevel:
+                    fail(location, "you must put your code inside a function")
```
Akuli, line 212: This error message is not nice if the problem is forgotten `global`.

For example, with `x: int` this in `foo.jou`, I get:

```
compiler error in file "foo.jou", line 1: you must put your code inside a function
```

Ideally the error would be e.g. `'global' keyword needed to create a global variable`

PR #710, opened 2025-01-27
Evaluate all if statements at compile time when possible

### LINKED ISSUE #497: Platform if statements inside class

### LINKED ISSUE #703: Platform if statements inside function

Akuli:
Fixes #497
Fixes #703


compiler/evaluate.jou
```diff
@@ -25,37 +25,60 @@ def get_special_constant(name: byte*) -> int:
             return -1
 
 
-def evaluate_condition(expr: AstExpression*) -> bool:
+def evaluate_condition(expr: AstExpression*) -> int:
```
Akuli, line 28:
```suggestion
def evaluate_condition(expr: AstExpression*) -> int:  # 1=true 0=false -1=error
```

compiler/evaluate.jou
```diff
@@ -25,37 +25,60 @@ def get_special_constant(name: byte*) -> int:
...
-                return True
-            fail(expr->location, "cannot evaluate condition at compile time")
+            return get_special_constant(expr->varname)
         case AstExpressionKind.And:
-            return evaluate_condition(&expr->operands[0]) and evaluate_condition(&expr->operands[1])
+            match evaluate_condition(&expr->operands[0]):
+                case 1:
+                    return evaluate_condition(&expr->operands[1])
+                case 0:
+                    return 0  # left side false, result false so don't evaluate right side
```
Akuli, line 37:
```suggestion
                    return 0  # left side false, don't evaluate right side
```

compiler/evaluate.jou
```diff
@@ -25,37 +25,60 @@ def get_special_constant(name: byte*) -> int:
...
+                case -1:
+                    return -1
+                case _:
+                    assert False
         case AstExpressionKind.Or:
-            return evaluate_condition(&expr->operands[0]) or evaluate_condition(&expr->operands[1])
+            match evaluate_condition(&expr->operands[0]):
+                case 1:
+                    return 1  # left side true, result true so don't evaluate right side
```
Akuli, line 45:
```suggestion
                    return 1  # left side true, don't evaluate right side
```

PR #713, opened 2025-01-27
Bootstrap script improvements

### LINKED ISSUE #711: Add caching to bootstrap.sh
Would be nice if I wouldn't have to redo the whole bootstrapping process from scratch when I e.g. delete `./jou_bootstrap` accidentally. Perhaps we could keep around all different versions in subdirectories.

Akuli:
- Save all bootstrapped compilers. Commit hash is included in folder name, so that this will work even if commits are removed from the list (unlikely).
- When a new commit is added to the bootstrap process, don't recompile everything from scratch.
- ~~For the first commit, use the compiler written in C directly instead of compiling a compiler written in Jou with it.~~ (failed on Windows for some reason)

Fixes #711

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -93,14 +93,15 @@ for i in ${!commits[@]}; do
         (cd $folder && unzip -q llvm_headers.zip)
     fi
 
-    if [ $i != 0 ]; then
+    if [ $i == 0 ]; then
+        # Only build the compiler written in C (jou_stage1.exe)
+        (cd $folder && make jou$exe_suffix)
+    else
         cp $(folder_of_commit $((i-1)))/jou$exe_suffix $folder/jou_bootstrap$exe_suffix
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -43,45 +43,67 @@ else
...
     show_message "Checking out and compiling commit ${commit:0:10} ($((i+1))/${#commits[@]})"
 
-    git checkout -q $commit
+    folder=$(folder_of_commit $i)
+    rm -rf $folder >/dev/null
+    mkdir -vp $folder
+
+    # If you know a better way to checkout a commit into a folder, let me know.
+    git archive --format=zip --output $folder/repo.zip $commit
+    (cd $folder && unzip -q repo.zip && rm repo.zip)
```
Akuli, line 82:
```suggestion
    # This seems to be the best way to checkout a commit into a folder.
    git archive --format=tar $commit | (cd $folder && tar xf -)
```

PR #719, opened 2025-02-02
Fix 'make clean'

Akuli:
Currently folder paths in `Makefile.*` and `bootstrap.sh` are not the same like they should be, which causes `make clean` to print a lot of unnecessary output.


Akuli:
Actually, I think `make clean`
- shouldn't delete `tmp/bootstrap_cache`
- should delete the `jou_bootstrap` executable

Currently, `make clean` doesn't delete `jou_bootstrap`, so running `make` again is pretty fast, but if the bootstrap script is changed, it will surprise you by rebootstrapping the whole thing.

PR #720, opened 2025-02-03
Detect undefined variables and return values with UVGs

### LINKED ISSUE #3: No compile error, if you forget return
```python
cdecl puts(string: byte*) -> int

def main() -> int:
    puts("Hello World")
    # No return statement. Should not compile, but currently segfaults at runtime.
```

I wonder if I should ask LLVM whether the end of the function can be reached without a `return` statement, or should I check for that myself.

### LINKED ISSUE #566: Undefined value warnings/errors

Akuli:
Fixes #3 (lol)
Fixes #566 

~~Marking as draft for now, because compiling the compiler with itself creates a lot of warnings from correct code. The problem is that `match` statement doesn't invoke UB if you match over an enum that is out of range, so the compiler assumes it can happen every time you `match` over an enum.~~


tests/should_succeed/return_none.jou
```diff
@@ -0,0 +1,18 @@
...
+
+def skipper(skip: bool, c: byte) -> None:
+    if skip:
+        return
+    putchar(c)
+    putchar('\n')
+
+def main() -> int:
+    skipper(False, 'a')  # Output: a
+    skipper(True, 'a')
```
Akuli, line 16:
```suggestion
    skipper(True, 'b')
```

compiler/builders/uvg_builder.jou
```diff
@@ -162,11 +179,16 @@ class UBuilder:
     # value as to
     def cast(self, value: int, to: Type*) -> int:
         if to->is_pointer_type():
+            # TODO: is this needed?
```
Akuli, line 182:
```suggestion
            # Needed to keep track of pointers to local variables, implicit casts are basically everywhere
```

compiler/builders/uvg_builder.jou
```diff
@@ -10,29 +9,37 @@ import "./ast_to_builder.jou"
...
+def FALSE_ID() -> int:
+    return -3
+
+
 class UBuilder:
     uvg: Uvg
     current_block: UvgBlock*
-    returns_a_value: bool
     location: Location
+    warn_unreachable: bool
```
Akuli, line 32: Do we really need this in this PR? This PR should only check undefined values, not unreachable code.

doc/compiler_internals/uvg.md
```diff
@@ -173,10 +173,12 @@ Also, all values may be undefined in the beginning of the start block.
 When we have computed the statuses of variables at the end of a block,
 and they differ from the previous analysis,
 we take the blocks that used the outdated statuses and queue them for another round of analyzing.
+The algorithm will terminate eventually as long as a cyclic change of statuses
+(e.g. `foo -> bar -> baz -> foo`) is not possible.
 
-At this point, we know which statuses are possible, and control flow has basically been taken care of.
-Because we only stored the statuses at the end of each block,
-we must loop through the instructions in each block again to show warnings.
+Once all that is done, we know which statuses are possible at the start and end of each block,
```
Akuli, line 179:
```suggestion
Once all that is done, we know which statuses are possible at the start and end of each block.
```

compiler/builders/uvg_builder.jou
Akuli: No

Issue #718, opened 2025-01-30
GitHub Actions: replace some actions with bash+curl?


Akuli:
In my experience, actions written/maintained by JavaScript people tend to bitrot easily. If I don't work on a project for a few years, the CI is almost always red because some action was deleted or changed in an incompatible way.

For example, Jou is currently using the latest version of `download-artifact` action and there's already a deprecation warning:

![Image](https://github.com/user-attachments/assets/10ccb2cf-529e-4d6c-8f09-ab2a54854c07)

All I really want is a couple HTTP requests to download files. To me, that sounds like a job for `bash` and `curl`, which I already use in various actions. Maybe I should replace more actions with bash+curl.

Akuli:
Turns out that this doesn't make sense for artifacts (#721), but it's still true we currently we use too many external actions:

```
$ grep -r ' uses:' .github | cut -d: -f3- | sort -u
 actions/checkout@v4
 actions/checkout@v4  # TODO: upgrade to v4 in other places
 actions/download-artifact@v4
 actions/github-script@v6
 actions/setup-python@v5
 actions/upload-artifact@v4
 cross-platform-actions/action@v0.26.0
 editorconfig-checker/action-editorconfig-checker@d4fca16fc71adef10fbe101903b654449fa9570c
 ncipollo/release-action@v1
 stefanzweifel/git-auto-commit-action@v4
```

I'm starting to think that anything starting with `actions/` is official and fine, but I should replace other actions.

PR #721, opened 2025-02-03
Stop using actions/download-artifact and actions/upload-artifact

Akuli:
Part of #718


Akuli:
Now `download-artifact` stuff is done. Let's do `upload-artifact` next.

Akuli:
Turns out there's no documented way to upload an artifact with the REST api, so this doesn't make sense.

PR #722, opened 2025-02-03
Do not use an action for editorconfig-checker

Akuli:
Part of #718

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -17,11 +17,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      # Hard-coded commit, because the latest changes aren't tagged.
-      - uses: editorconfig-checker/action-editorconfig-checker@d4fca16fc71adef10fbe101903b654449fa9570c
-        with:
-          version: 2.7.0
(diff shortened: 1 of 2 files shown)


.github/workflows/checks.yml
```diff
@@ -17,11 +17,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      # Hard-coded commit, because the latest changes aren't tagged.
-      - uses: editorconfig-checker/action-editorconfig-checker@d4fca16fc71adef10fbe101903b654449fa9570c
-        with:
-          version: 2.7.0
-      - run: editorconfig-checker
+      - run: wget https://github.com/editorconfig-checker/editorconfig-checker/releases/download/v3.2.0/ec-linux-amd64.tar.gz
+      - run: tar xvf ec-linux-amd64.tar.gz bin/ec-linux-amd64
```
Akuli, line 21: The verbose flag doesn't really add much here

```suggestion
      - run: tar xf ec-linux-amd64.tar.gz bin/ec-linux-amd64
```

Akuli:
I always like when PR deletes more lines than adds while solving a problem :)

PR #736, opened 2025-02-05
Support True and False in compile-time if statements

### LINKED ISSUE #730: True and False are not recognized as compile-time constants
```python
import "stdlib/io.jou"

if False:
    def foo() -> int:
        pass

def main() -> int:
    printf("hi\n")
    return 0
```

Expected output: `hi`

Actual output: `compiler error in file "foo.jou", line 3: cannot evaluate condition at compile time`

This feature would be similar to `#if 0 ... #endif` in C: even though C has multi-line comments (Jou doesn't), people still find it useful to `#if 0` instead of a comment sometimes.

There is currently a weird workaround:

https://github.com/Akuli/jou/blob/839707c24d838b7cc76e2b06fcbb3db7358607a8/compiler/uvg_analyze.jou#L18-L19

Akuli:
Fixes #730

### DIFF
--- a/compiler/uvg_analyze.jou
+++ b/compiler/uvg_analyze.jou
@@ -15,7 +15,7 @@ enum VarStatus:
     DontAnalyze         # The "don't analyze" UVG instruction has been used.
 
 
-# TODO: "if False:" doesn't work
+# TODO: use "if False:" once #736 is merged
 if WINDOWS and not WINDOWS:
     def print_statuses(uvg: Uvg*, statuses: VarStatus*) -> None:
         printf("VAR STATUSES:")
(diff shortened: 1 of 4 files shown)


compiler/uvg_analyze.jou
```diff
@@ -15,7 +15,7 @@ enum VarStatus:
     DontAnalyze         # The "don't analyze" UVG instruction has been used.
 
 
-# TODO: "if False:" doesn't work
+# TODO: use "if False:" once #736 is merged
```
Akuli, line 18:
```suggestion
# TODO: use "if False:"
```

PR #739, opened 2025-02-05
Show warning if nested loops use the same variable

### LINKED ISSUE #362: Warning for reusing for loop variable
I have multiple times written Jou code like this:

```
for i = 0; i < 10; i++:
    ...
    for i = 0; i < 10; i++:
        ...
```

The solution is to rename the variable in one of the loops, but it would be nice if the compiler warned about this.

I'm not sure how I should detect loops like this. Maybe just detect any two nested for loops that start with `for foo = bar; ...`, where `foo` parts are the same variable? I don't want to make it too general. For example, this should not trigger the warning:

```
for i = 0; i < 10; i++:
    ...

for i = 0; i < 10; i++:
    ...
```

Akuli:
Fixes #362

### DIFF
--- a/compiler/parser.jou
+++ b/compiler/parser.jou
@@ -868,7 +868,7 @@ class Parser:
     def enter_loop_body(self, varname: byte*, location: Location) -> None:
         if varname != NULL:
             for i = 0; i < self->nloops; i++:
-                if strcmp(self->loop_vars[i], varname) == 0:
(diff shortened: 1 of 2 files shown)


compiler/parser.jou
```diff
@@ -864,14 +865,30 @@ class Parser:
             else_body = else_body,
         }
 
+    def enter_loop_body(self, loop_var_token: Token*) -> None:
```
Akuli, line 868: This function should really just take a string and a location

PR #751, opened 2025-02-07
Make global variable definitions private by default

Akuli:
You can use `@public` to make a global variable accessible from other files (just like functions):

```python
@public
global foo: int
```

Without `global`, the variable is private to the file (known as `static` in C and similar languages). Other files cannot access it directly even if they import the file.

Part of #84

### DIFF
--- a/compiler/builders/llvm_builder.jou
+++ b/compiler/builders/llvm_builder.jou
@@ -570,13 +570,23 @@ def build_llvm_ir(ast: AstFile*) -> LLVMModule*:
     builder_wrapper = EitherBuilder{lbuilder = &builder}
 
     for g = ast->types.globals; g < &ast->types.globals[ast->types.nglobals]; g++:
-        t = type_to_llvm(g->type)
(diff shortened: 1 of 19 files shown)


compiler/builders/llvm_builder.jou
```diff
@@ -570,13 +570,23 @@ def build_llvm_ir(ast: AstFile*) -> LLVMModule*:
...
         match stmt->kind:
+            case AstStatementKind.GlobalVariableDef:
+                globalptr = LLVMGetNamedGlobal(module, stmt->global_var_def.name)
+                assert globalptr != NULL
+                # TODO: better way to get the type
+                vartype: Type* = NULL
+                for g = ast->types.globals; g < &ast->types.globals[ast->types.nglobals]; g++:
+                    if strcmp(g->name, stmt->global_var_def.name) == 0:
+                        vartype = g->type
+                        break
```
Akuli, line 585:
```suggestion
                vartype = ast->types.find_global_var(stmt->global_var_def.name)
```

PR #762, opened 2025-02-11
Delete broken_tests/simplify_cfg.c

Akuli:
This is the only remaining C file in the whole project. I left it around in case it would be useful when implementing UVGs, but it is no longer needed, because I already implemented UVGs.


Akuli:
I would like to merge #763 first.

PR #766, opened 2025-02-12
Fix a corner case in compile-time if statement handling

Akuli:
Needed for #763 

Previously, compile-time evaluatable if statements inside runtime if statements were never evaluated.

For example, if you don't use windows, the following would procude a compiler warning even though the code would run on Windows:

```python
if some_dynamic_thing() == other_dynamic_thing():
    if WINDOWS:
        do_stuff()  # Warning: this code will never run
```

This PR removes the warning, because it evalutes `if WINDOWS:` at compile time before we do the warnings.

### DIFF
--- a/tests/should_succeed/compile_time_if.jou
+++ b/tests/should_succeed/compile_time_if.jou
@@ -1,4 +1,5 @@
 import "stdlib/io.jou"
+import "stdlib/process.jou"
 
 # TODO: "if WINDOWS" doesn't work very well inside functions yet.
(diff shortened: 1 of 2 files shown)


tests/should_succeed/compile_time_if.jou
```diff
@@ -56,4 +57,13 @@ if True:
         if False:
             printf("hello\n")
```
Akuli, line 58:
```suggestion
            totally(not valid, at.all)
```

PR #772, opened 2025-02-18
Declare the `asprintf()` function

Akuli:
This function is like `sprintf()`, except that it allocates a new string of the right size with `malloc()`. It is not in the C standard, but all 4 supported platforms seem to have it.

Thank you @taahol for letting me know that this function exists :)


Akuli:
I ran MacOS CI manually and it seems to be succeeding.

PR #774, opened 2025-02-19
Add support for LLVM 17, 18 and 19

Akuli:

PR #777, opened 2025-02-19
Fix NetBSD support

Akuli:
A PR for changes done by @taahol in #774. Depends on #774, and I will merge it first.


taahol:
Don’t merge yet; I have some tricks in mind.

PR #781, opened 2025-02-24
Resolve relative paths in flags given with `link` keyword

Akuli:
You can now do `link "../libfoo.a"` in a Jou file. The compiler will pass `/full/path/to/libfoo.a` to the linker. When determining the full path, the compiler interprets `"../libfoo.a"` as relative to the directory where the Jou file is, just like with imports. This also works if the path is not at the start of the `link` string, e.g. `link "-L../my_folder"`.

I'm not sure if this is too magical for Jou, but it seems very nice and practical because it doesn't matter which directory you run Jou files from. You can do `jou folder/filename.jou`, or `cd folder` and then `jou filename.jou`. It will "just work".


compiler/parser.jou
```diff
@@ -244,7 +244,7 @@ class Parser:
         self->eat_newline()
 
         # TODO: replace "./" and "../" with actual folder paths
```
Akuli, line 246: done
```suggestion
```

compiler/paths.jou
```diff
@@ -244,3 +245,119 @@ def simplify_path(path: byte*) -> None:
...
+        if is_ascii_whitespace(*p) and quoting:
+            # add quote before whatever thing we added that needs quoting
+            insert(&p[1], quote)
+            quoting = False
+
+        if not starts_with(p, "./") and not starts_with(p, "../"):
+            continue
+
+        # If we have "../" do nothing when we see the "./" part
+        if p > result and p[-1] == '.':
```
Akuli, line 298: This will not work if the user writes `.../foo` (very unlikely). Let's fix it anyway:
```suggestion
        if p > result and starts_with(&p[-1], "../"):
```

runtests.sh
```diff
@@ -254,15 +259,15 @@ function run_test()
 counter=0
 skipped=0
 
-for joufile in examples/*.jou examples/aoc*/day*/part*.jou tests/*/*.jou tests/should_succeed/double_dotdot_import/*/*.jou; do
+for joufile in examples/*.jou examples/aoc*/day*/part*.jou tests/*/*.jou tests/should_succeed/double_dotdot_import/*/*.jou compiler/paths.jou; do
```
Akuli, line 262:
```suggestion
for joufile in examples/*.jou examples/aoc*/day*/part*.jou tests/*/*.jou tests/should_succeed/double_dotdot_import/*/*.jou; do
```

runtests.sh
```diff
@@ -254,15 +259,15 @@ function run_test()
...
     if ! [[ $joufile == *"$file_filter"* ]]; then
         # Skip silently, without showing that this is skipped.
         # This produces less noisy output when you select only a few tests.
         continue
     fi
 
     case $joufile in
-        examples/* | tests/should_succeed/*) correct_exit_code=0; ;;
+        examples/* | tests/should_succeed/* | compiler/*) correct_exit_code=0; ;;
```
Akuli, line 270:
```suggestion
        examples/* | tests/should_succeed/*) correct_exit_code=0; ;;
```

PR #796, opened 2025-03-18
Fix bug when parsing files that don't have a trailing newline

### LINKED ISSUE #795: Trailing newline needed to compile
Jou programs do not compile if the last line has indentation but no trailing newline character (\n) at the end.

Example (in bash):

```bash
akuli@akuli-desktop:~/jou$ printf 'import "stdlib/io.jou"\ndef main() -> int:\n    return 0' > a.jou
akuli@akuli-desktop:~/jou$ jou a.jou
compiler error in file "a.jou", line 3: expected end of line, got less indentation
```

Akuli:
Fixes #795


tests/should_succeed/eof_no_newline.jou
```diff
@@ -0,0 +1,2 @@
+def main() -> int:
```
Akuli, line 1:
```suggestion
import "stdlib/io.jou"

def main() -> int:
    printf("Hi\n")  # Output: Hi
```

Akuli:
CI failure is unrelated. Somehow downloading mingw became really slow (about 30% downloaded in 5 minutes).

Issue #791, opened 2025-03-15
Inline functions


Akuli:
I want to add an `@inline` decorator that causes the Jou compiler to inline a function. This is very different from most other  compilers, where inlining is only a hint and you never know whether the compiler decides to actually inline the function or not.

For example, this:

```python
@inline
def add(x: int, y: int) -> int:
    return x + y

def main() -> int:
    printf("%d\n", add(1, 2))  # Output: 3
```

would be same as this:

```python
def main() -> int:
    printf("%d\n", 1 + 2)  # Output: 3
```

We need to be careful about variables inside the inlined function. They must not be mixed with variables in the outer function when compiling. For example, if `main()` function above has a variable named `x`, that should not affect anything.

This is important mostly because inline functions [are needed for the `.append()` method of a generic `List`](https://github.com/Akuli/jou/issues/527#issuecomment-2537443264).

PR #797, opened 2025-03-19
Add support for global compile-time constants

### LINKED ISSUE #369: Global constants
In C you would use `#define` to create constants. What should be the recommended way in Jou? Should we add new syntax, something like `globalconstant FOO = 1`?

I think currently the best you can do is a function that returns the value, because global variables cannot have an initial value yet. This is actually what `config.jou` does (see the Makefiles).

Akuli:
Syntax:

```python3
# Private constant
const FOO: int = 123

# Public constant
@public
const BAR: int = 456
```

This is an initial, somewhat simple implementation, and doesn't support all the things people might expect global constants to support (e.g. simple arithmetic or using global constants to specify array sizes). This can be expanded later as needed.

Fixes #369


compiler/ast.jou
```diff
@@ -208,7 +210,8 @@ class AstExpression:
         array: AstArray
         varname: byte[100]
         float_or_double_text: byte[100]
-        operands: AstExpression*  # Only for operators. Length is arity, see get_arity()
+        operands: AstExpression*
```
Akuli, line 213: preserve comment
```suggestion
        operands: AstExpression*  # Only for operators. Length is arity, see get_arity()
```

compiler/ast.jou
```diff
@@ -316,6 +319,8 @@ class AstExpression:
                 printf("and\n")
             case AstExpressionKind.Or:
                 printf("or\n")
+            case AstExpressionKind.Constant:
+                printf("constant\n")
```
Akuli, line 323: Should be moved up just after `GetEnumMember`, should print the value of the constant

Moosems:
Would this work and give a compile time constant?
```python
@public
const BAR: int = foo(456)
```

Akuli:
No, only simple values will be supported in compile-time constants. However, you can achieve the same result in a couple ways.

You could make a function that computes and returns `foo(456)`. The optimizer will attempt to simplify your code and probably ends up baking a hard-coded value into the executable, especially if `foo` is decorated with `@inline` (#791).

```python3
def bar() -> int:
    return foo(456)
```

You could hard-code the value of `BAR` and assert that it is correct when the program starts.

```python3
const BAR = 123456

def main() -> int:
    assert foo(123) == BAR
    ...
```

PR #802, opened 2025-03-21
Add support for `const` strings

### LINKED ISSUE #800: `const` strings
```python3
import "stdlib/io.jou"

const a: byte* = "hi"

def main() -> int:
    printf("%s\n", a)
    return 0
```

Expected result: no error
Actual result: `compiler error in file "a.jou", line 3: cannot evaluate value of constant at compile time`

Akuli:
Example:

```python3
const MESSAGE: byte* = "hello"
```

Implemented by reusing `AstExpressionType.Constant` for strings, extending the idea of #801. I also applied it to `NULL`.

The following do not work yet, because as it is now, no implicit casts are applied in `const` statements:

```python3
const MESSAGE2: byte* = NULL
const MESSAGE3: byte[50] = "hello there"
```

Fixes #800


compiler/builders/ast_to_builder.jou
```diff
@@ -256,6 +256,11 @@ class AstToBuilder:
 
     def build_constant(self, constant: Constant*) -> EitherBuilderValue:
         match constant->kind:
+            case ConstantKind.ArrayString:
+                # TODO: rename the builder methods according to ConstantKind enum
```
Akuli: Might as well do it right away

Issue #455, opened 2023-12-15
Compile time variables and parameters


Moosems:
Similar to how Mojo does it, what do you think of compile time variables and separating parameters (compile time) and arguments (runtime) (see [here](https://docs.modular.com/mojo/manual/basics.html#parameterization))?

Akuli:
Duplicate of #369

Moosems:
I wouldn't group this with constants...

Akuli:
I see. This is basically what we discussed in #257.

Akuli:
IMO the right solution is to allow accessing environment variables during compile time. Environment variables would be used to decide and configure what gets compiled ("parameters"), and command-line arguments would be passed when running the program ("arguments").

For example, let's say you want to make a hello world program where the greeting (e.g. `Hello %s` or `Hi there, %s`) can be configured at compile time, but the target of the greeting (`%s`) is given at runtime. You would do this:

```python
import "stdlib/io.jou"

def main(argc: int, argv: byte**) -> int:
    format = compile_time_getenv("FORMAT")  # compile-time parameter
    printf(format, argv[1])  # runtime argument
    return 0
```

Then you would pass environment variables when compiling the program:

```
FORMAT="Hello %s" jou -o program program.jou
```

And run it like this:

```
./program world
```

This would print `Hello world`.

And once #257 is implemented, you could compile and run at once:

```
FORMAT="Hello %s" jou program.jou world
```

Here's one situation where accessing env vars during compile time is extremely useful. Let's say you have a project that for whatever reason invokes the Jou compiler through a script. The script is messy and you really don't want to edit it for whatever reason (e.g. it could be auto-generated by a stupid tool such as `cmake`). With env vars, you could **still** pass parameters to the Jou compiler through environment variables, because each child process inherits the environment of its parent.

Moosems:
At compile time if I have a hard-coded integer, is that baked into the executable?

Akuli:
Yes, and whatever `compile_time_getenv()` returns would also be baked into the executable.

Akuli:
No, `compile_time_getenv()` doesn't exist yet.

PR #803, opened 2025-03-21
Compiler refactor: use `const` when appropriate

Akuli:


Akuli:
Perhaps `JOU_CLANG_PATH` could be an environment variable, as in #455.

PR #805, opened 2025-03-26
Implement generics

### LINKED ISSUE #527: Generics
I couldn't find anything on generics, is that a language concept yet? Generally something that happens later though

Akuli:
**Jou will finally have generics!!!!**

Simple example:

```python3
import "stdlib/io.jou"

class Pair[T]:
    x: T
    y: T

def main() -> int:
    point = Pair[int]{x=1, y=2}
    printf("%d %d\n", point.x, point.y)  # Output: 1 2
    return 0
```

Complex example: (a generic implementation of a list that grows automatically)

```python3
import "stdlib/io.jou"
import "stdlib/mem.jou"

class List[T]:
    ptr: T*
    len: long
    alloc: long
    last_size_of_item: long

    def end(self) -> T*:
        ptr = self->ptr as byte*
        return &ptr[self->len * self->last_size_of_item] as T*

    def grow(self, num_bytes_needed: long) -> None:
        if self->alloc >= num_bytes_needed:
            return

        if self->alloc == 0:
            self->alloc = 10

        while self->alloc < num_bytes_needed:
            self->alloc *= 2

        self->ptr = realloc(self->ptr, num_bytes_needed)
        assert self->ptr != NULL

    def append(self, item: T*, size_of_item: long) -> None:
        if self->last_size_of_item == 0:
            self->last_size_of_item = size_of_item
        else:
            assert self->last_size_of_item == size_of_item

        self->grow((self->len + 1) * size_of_item)
        memcpy(self->end(), item, size_of_item)
        self->len++

def main() -> int:
    nums = List[int]{}
    n = 6
    nums.append(&n, sizeof(n))
    n++
    nums.append(&n, sizeof(n))
    n++
    nums.append(&n, sizeof(n))

    # Output: 6
    # Output: 7
    # Output: 8
    for p = nums.ptr; p < nums.end(); p++:
        printf("%d\n", *p)

    free(nums.ptr)
    return 0
```

Unfortunately, you still need to pass the size of the item when appending: `nums.append(&foo, sizeof(foo))` instead of a simple `nums.append(foo)`. The reason is that the compiler doesn't know the size of the item when it compiles methods in the class. This can later be worked around with `@inline` (#791).

It is possible to make the compiler crash in various corner cases. For example, the compiler crashes if it needs to convert the unknown `T` typevar into an LLVM type. I will create issues and fixs the crashes when they come up in practice.

Fixes #527


compiler/parser.jou
```diff
@@ -298,13 +334,42 @@ class Parser:
                 self->tokens++
 
                 result = AstType{
-                    location = location,
                     kind = AstTypeKind.Array,
```
Akuli, line 337: Preserve it
```suggestion
                    location = location,
                    kind = AstTypeKind.Array,
```

compiler/parser.jou
```diff
@@ -298,13 +334,42 @@ class Parser:
...
+                while not self->tokens->is_operator("]"):
+                    params = realloc(params, sizeof(params[0]) * (nparams + 1))
+                    params[nparams++] = self->parse_type()
+                    if not self->tokens->is_operator(","):
+                        break
+                    self->tokens++
+
+                if nparams == 0:
+                    fail(self->tokens->location, "empty [] is not valid syntax when specifying a type")
+
```
Akuli, line 358: Let's just let it fail later

```suggestion
```

compiler/typecheck/step2_populate_types.jou
```diff
@@ -164,20 +164,32 @@ def check_class_member_not_exist(t: Type*, name: byte*, location: Location) -> N
 
 
 def handle_class_members(ft: FileTypes*, classdef: AstClassDef*) -> None:
-    # Previous type-checking step created an opaque class.
+    # Previous type-checking step created a temporary type. It now becomes a class.
```
Akuli, line 167: No change was intended here

```suggestion
    # Previous type-checking step created an opaque class.
```

compiler/types.jou
```diff
@@ -301,6 +418,31 @@ def create_enum(name: byte*, membercount: int, membernames: byte[100]*) -> Type*
...
+    class_name_len = strcspn(class_name, "[")
+
+    assert nparams >= 1
+    total_len = class_name_len + strlen("[]") + (nparams - 1)*strlen(", ")
+    for i = 0; i < nparams; i++:
+        total_len += strlen(params[i]->name)
+
+    result: byte[500]
+    if total_len < sizeof(result):
+        snprintf(result, sizeof(result), "%.*s[", class_name_len as int, class_name)
```
Akuli, line 434: We know it fits
```suggestion
        sprintf(result, "%.*s[", class_name_len as int, class_name)
```

PR #808, opened 2025-03-26
Compiler code cleanup: delete TypeKind.OpaqueClass

Akuli:
This was useless. It felt like a good idea when it was added but it really serves no purpose.

### DIFF
--- a/compiler/typecheck/step2_populate_types.jou
+++ b/compiler/typecheck/step2_populate_types.jou
@@ -167,7 +167,7 @@ def handle_class_members(ft: FileTypes*, classdef: AstClassDef*) -> None:
     # Previous type-checking step created an opaque class.
     t = classdef->type
     assert t != NULL
-    assert t->kind == TypeKind.OpaqueClass
+    assert t->kind == TypeKind.Class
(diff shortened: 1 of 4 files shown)


compiler/typecheck/step2_populate_types.jou
```diff
@@ -167,7 +167,7 @@ def handle_class_members(ft: FileTypes*, classdef: AstClassDef*) -> None:
     # Previous type-checking step created an opaque class.
     t = classdef->type
     assert t != NULL
-    assert t->kind == TypeKind.OpaqueClass
+    assert t->kind == TypeKind.Class
     t->kind = TypeKind.Class
```
Akuli, line 171:
```suggestion
```

compiler/typecheck/step2_populate_types.jou
```diff
@@ -167,10 +167,7 @@ def handle_class_members(ft: FileTypes*, classdef: AstClassDef*) -> None:
     # Previous type-checking step created an opaque class.
```
Akuli, line 167:
```suggestion
    # Previous type-checking step created an empty class.
```

PR #810, opened 2025-03-26
Fix bugs with generic classes

Akuli:
- Previously, generic classes imported from other files behaved like non-generic classes.
- It is now possible to use `GenericClass[Foo]` in method signatures and fields of other classes. Previously that caused the type `GenericClass[Foo]` to have no fields or methods due to a problem with caching the type.


compiler/typecheck/step1_create_types.jou
```diff
@@ -18,6 +18,21 @@ import "../types.jou"
 import "./common.jou"
 
 
+# When defining a generic class Foo[T], this function creates the T and
+# attaches it to the Foo type. It also renames the type from "Foo" to "Foo[T]".
+def handle_generics(t: Type*, classdef: AstClassDef*) -> None:
+    n = classdef->n_generic_typevars  # number of generic parameters, e.g. 2 for Foo[T1, T2]
```
Akuli, line 24:
```suggestion
    assert t->kind == TypeKind.Class
    n = classdef->n_generic_typevars  # number of generic parameters, e.g. 2 for Foo[T1, T2]
```

PR #816, opened 2025-03-29
Use lists in examples/aoc2023 and examples/aoc2024

Akuli:
The goal is not to rewrite everything possible with lists, but at least get rid of the `class List: ...` that I rewrote every time I needed a list.


Akuli:
I grepped for `realloc`, turns out there's more work than I expected :)

PR #821, opened 2025-04-02
Add initial support for `@inline`

### LINKED ISSUE #791: Inline functions

Akuli:
With this PR, the compiler fully supports the following:
```python
class List[T]:
    ...
    @inline
    def append(self, item: T) -> None:
        self->grow(self->len + 1, sizeof(item))
        *self->end() = item
        self->len++
```

This means that we can finally have a simple `list.append(item)`. (That will be implemented in another PR.)

While `@inline` seems to work for this use case, there are probably several corner cases where it fails. I will report them as GitHub issues when I discover them.

The code is also somewhat spaghetti. The biggest problem is that type checking the content of class `Foo[T]` leaves type `T` into the AST, and when we inline the method, the compiler must use a concrete type such as `int` every time it sees `T`.

Fixes #791


compiler/builders/ast_to_builder.jou
```diff
@@ -22,14 +22,39 @@ class Loop:
...
+
 class AstToBuilder:
     builder: EitherBuilder*
     locals: List[LocalVar]
     loops: List[Loop]
-    returns_a_value: bool
 
     location: Location
 
+    # If not NULL, this builder is used to handle calling an @inline function
```
Akuli, line 42:
```suggestion
    # If not NULL, this builder is being used to handle calling an @inline function
```

compiler/builders/ast_to_builder.jou
```diff
@@ -604,14 +704,23 @@ class AstToBuilder:
 
 
 @public
-def feed_ast_to_builder(func_ast: AstFunctionOrMethod*, func_location: Location, is_main_file: bool, builder: EitherBuilder*) -> None:
+def feed_ast_to_builder(
+    func_ast: AstFunctionOrMethod*,
+    func_location: Location,
+    is_main_file: bool,
+    builder: EitherBuilder*,
+) -> None:
```
Akuli, line 712: Let's preserve formatting:
```suggestion
def feed_ast_to_builder(func_ast: AstFunctionOrMethod*, func_location: Location, is_main_file: bool, builder: EitherBuilder*) -> None:
```

doc/inline.md
```diff
@@ -0,0 +1,112 @@
+# Inline functions
+
+Jou functions can be marked with the `@inline` decorator.
```
Akuli, line 3:
```suggestion
Jou functions and methods can be decorated with the `@inline` decorator.
```

tests/should_succeed/inline.jou
```diff
@@ -0,0 +1,30 @@
+import "stdlib/io.jou"
+
+class Foo[T]:
+    x: T
+
+    # This wouldn't work without @inline because size of stack
+    # variable new_x and return value would be unknown.
```
Akuli, line 7:
```suggestion
    # variable new_x and return value would be unknown.
    #
    # TODO: add a test for this, ensure compiler gives a reasonable error message
```

PR #827, opened 2025-04-02
Fix calling generic inline methods through self

### LINKED ISSUE #826: Calling inline method through self crashes the compiler
The following crashes the compiler:

```python
class Foo[T]:
    @inline
    def inline_method(self) -> None:
        pass

    def foo(self) -> None:
        self->inline_method()


def main() -> int:
    return 0
```

Expected result: no output
Actual result: `Assertion 'from[i] != to[i]' failed in file "compiler/types.jou", line 125.`

This is a bug left behind by the lazy/spaghetti implementation in #821. There are probably many more.

I discovered this while trying to improve `List.extend_from_ptr()`.

Akuli:
Fixes #826


compiler/types.jou
```diff
@@ -133,14 +131,22 @@ class Type:
...
-                if not self->classdata.is_generic():
-                    # Do nothing for non-generic classes
-                    return self
-
                 generic_params = List[Type*]{}
                 for p = self->classdata.generic_params.ptr; p < self->classdata.generic_params.end(); p++:
                     generic_params.append((*p)->substitute_generic_params(from, to, n))
 
+                # Special case: No substitutions need to be done.
+                # Used e.g. when class List[T] references to itself as List[T].
```
Akuli, line 139:
```suggestion
                # Used e.g. when List[T] occurs inside the body of class List[T].
```

PR #828, opened 2025-04-02
Return the removed element in `list.pop()`

Akuli:
Previously, `List.pop()` returned nothing due to limitations of the Jou language/compiler. With `@inline`, it is now possible to returned the value removed from the end of the list, just like Python does.

### DIFF
--- a/stdlib/list.jou
+++ b/stdlib/list.jou
@@ -69,7 +69,18 @@ class List[T]:
         end = &start[self->len * self->itemsize]
         return end as T*
 
-    # TODO: It is currently not possible to return the deleted element unless it is a pointer.
-    def pop(self) -> None:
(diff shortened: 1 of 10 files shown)


stdlib/list.jou
```diff
@@ -69,7 +69,18 @@ class List[T]:
...
-    def pop(self) -> None:
+    # Return the last item from the list and return it.
+    #
+    # To remove an item at index i, you can replace it with the result of
+    # popping:
+    #
+    #    list.ptr[i] = list.pop()
+    #
+    # This works even if `i` refers to the last item of the list. In that case,
+    # the item is assigned back to the place where it was just "removed" from.
```
Akuli, line 80:
```suggestion
```

PR #834, opened 2025-04-03
stdlib/ascii.jou: Return `List[byte*]` instead of `byte**` from `split_by_ascii_whitespace()`

Akuli:


examples/aoc2023/day08/part2.jou
```diff
@@ -131,7 +131,7 @@ def get_z_counts(input: Input*, start: Node*) -> ZCounts:
                 # from here it cycles
                 non_repeating = List[long]{}
                 first_cycle = List[long]{}
-                for k = 0L; k < nstates; k++:
+                for k = 0; k < nstates; k++:
```
Akuli, line 134: no need to change this

```suggestion
                for k = 0L; k < nstates; k++:
```

PR #840, opened 2025-04-04
Fail CI if using ```python3 accidentally

Akuli:
Jou code in markdown files uses "Python" syntax highlighting. GitHub supports multiple ways of specifying it, at least the following:

    ```python
    ```python3
    ```py
    ```py3

`doctest.sh` only supports the `python` marker. This PR adds a check to CI that fails if any of the other markers are used.


Akuli:
Now it's failing with a good looking error message

screenshot-1743807864

PR #842, opened 2025-04-05
More bug fixes for inline methods in generic classes

### LINKED ISSUE #841: Cannot call inline function from another inline function in generic class
```python3
import "stdlib/io.jou"

class Foo[T]:
    @inline
    def bar(self, x: T) -> T:
        return x
    @inline
    def baz(self, x: T) -> T:
        return self->bar(x)

def main() -> int:
    f = Foo[int]{}
    printf("%d\n", f.bar(1))  # Output: 1
    printf("%d\n", f.baz(1))  # Output: 1
    return 0
```

Expected result: `1` printed twice
Actual result: `Assertion 'False' failed in file "compiler/builders/llvm_builder.jou", line 102.`

Akuli:
Fixes #841

### DIFF
--- a/compiler/builders/ast_to_builder.jou
+++ b/compiler/builders/ast_to_builder.jou
@@ -123,7 +123,7 @@ class AstToBuilder:
             # When inlining a generic method List[int].append(), map T to int
             assert args.len >= 1
             ast_self_class = ast->types.signature.get_self_class()
-            passed_self_class = self->map_type(sig->get_self_class())
+            passed_self_class = sig->get_self_class()
 
             assert ast_self_class->kind == TypeKind.Class
             assert passed_self_class->kind == TypeKind.Class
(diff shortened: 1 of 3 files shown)


compiler/builders/ast_to_builder.jou
```diff
@@ -117,13 +117,13 @@ class AstToBuilder:
...
         if sig->return_type != NULL:
-            info.return_value_ptr = self->builder->stack_alloc(sig->return_type, NULL)
+            info.return_value_ptr = self->builder->stack_alloc(self->map_type(sig->return_type), NULL)
 
         if sig->get_self_class() != NULL and sig->get_self_class()->classdata.is_generic():
             # When inlining a generic method List[int].append(), map T to int
             assert args.len >= 1
             ast_self_class = ast->types.signature.get_self_class()
-            passed_self_class = sig->get_self_class()
+            passed_self_class = self->map_type(sig->get_self_class())
```
Akuli, line 126: Probably not needed, because the method signature is mapped already

```suggestion
            passed_self_class = sig->get_self_class()
```

Akuli: Don't know why, but this is needed :)

PR #843, opened 2025-04-05
Simplify list methods

Akuli:
The `List` class contained various workarounds for compiler bugs that no longer exist. Clean them up.

The only user-facing change is that `List.grow()` and `List.extend_from_ptr()` now take one less parameters:

```python3
list = List[int]{}
arr = [1, 2, 3, 4, 5]

# Old
list.grow(500, sizeof(list.ptr[0]))
list.extend_from_ptr(arr, 5, sizeof(arr[0]))

# New
list.grow(500)
list.extend_from_ptr(arr, 5)
```

### DIFF
--- a/doc/lists.md
+++ b/doc/lists.md
@@ -177,7 +177,7 @@ Here `realloc()` is the function that the `List` class uses to allocate heap mem
 Unfortunately, the error message says nothing about `List`,
 because the `List` class uses a lot of [`@inline`](inline.md).
 You can also see that the list allocated 512 bytes
-even though 400 bytes would be enough for 100 ints. Each int is 4 bytes.
+even though 400 bytes would be enough for 100 ints (each `int` is 4 bytes).
 
 See [the UB documentation](ub.md#crashing-and-valgrind) for more about `jou --valgrind`.
 
--- a/doc/lists.md
+++ b/doc/lists.md
@@ -1,0 +1,459 @@
+# Lists
+
+TL;DR:
+
+```python
+import "stdlib/io.jou"
+import "stdlib/list.jou"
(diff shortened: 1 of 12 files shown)


doc/lists.md
```diff
@@ -150,23 +150,35 @@ and it is your responsibility to free that memory when you no longer need the li
...
+==25607==    at 0x484682F: realloc (vg_replace_malloc.c:1437)
+==25607==    by 0x1091EA: main (in /home/akuli/jou/jou_compiled/a/a)
+==25607==
 ```
 
+Here `realloc()` is the function that the `List` class uses to allocate heap memory.
+Unfortunately, the error message says nothing about `List`,
+because the `List` class uses a lot of [`@inline`](inline.md).
+You can also see that the list allocated 512 bytes
+even though 400 bytes would be enough for 100 ints. Each int is 4 bytes.
```
Akuli, line 180:
```suggestion
even though 400 bytes would be enough for 100 ints (each `int` is 4 bytes).
```

stdlib/list.jou
```diff
@@ -9,65 +9,50 @@ import "stdlib/mem.jou"
 class List[T]:
     ptr: T*         # Pointer to first item of the list (may change when list grows)
     len: long       # How many items are in the list
-    alloc: long     # How many bytes of space starting at ptr are allocated
-    itemsize: long  # This is either zero or sizeof(T). Set when appending.
+    alloc: long     # How many items fit into the list without needing to allocate more memory
```
Akuli, line 12:
```suggestion
    alloc: long     # How many items would fit into the allocated memory
```

Akuli:
This can't be merged yet due to #844.

PR #847, opened 2025-04-07
Reuse previously compiled object files to speed up compiling

### LINKED ISSUE #806: Same file name in different directories
Create file `math.jou` with the following:

```python3
import "stdlib/io.jou"
import "stdlib/math.jou"

def main() -> int:
    printf("%d\n", max(1, 2))
    return 0
```

Expected output: `2`
Actual output:

```
/usr/bin/ld: /home/akuli/jou/jou_compiled/math/math.o: in function `min':
math.jou:(.text+0x0): multiple definition of `min'; /home/akuli/jou/jou_compiled/math/math.o:math.jou:(.text+0x0): first defined here
/usr/bin/ld: /home/akuli/jou/jou_compiled/math/math.o: in function `max':
math.jou:(.text+0x10): multiple definition of `max'; /home/akuli/jou/jou_compiled/math/math.o:math.jou:(.text+0x10): first defined here
/usr/bin/ld: /home/akuli/jou/jou_compiled/math/math.o: in function `llmin':
math.jou:(.text+0x20): multiple definition of `llmin'; /home/akuli/jou/jou_compiled/math/math.o:math.jou:(.text+0x20): first defined here
/usr/bin/ld: /home/akuli/jou/jou_compiled/math/math.o: in function `llmax':
math.jou:(.text+0x30): multiple definition of `llmax'; /home/akuli/jou/jou_compiled/math/math.o:math.jou:(.text+0x30): first defined here
/usr/bin/ld: /lib/x86_64-linux-gnu/Scrt1.o: in function `_start':
(.text+0x17): undefined reference to `main'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

The problem is that both `math.jou` files (our file and stdlib/math.jou) are compiled to `jou_compiled/math.o`. They should be complied to two differently named object files to avoid one overwriting the other.

### LINKED ISSUE #823: Caching compilation results
On my computer, it takes 8 seconds for the Jou compiler to compile itself. This is not too bad, but the problem is that it takes 8 seconds even if I recomple after change a single character that is inside a comment.

In other words, Jou always recompiles the whole project, not just the changed parts.

Ideally, the compilation would be split up into a dependency graph consisting of steps whose outputs would be cached to `jou_compiled/`. This is how other compilers do it. In C, there is a build system (traditionally `make`) for doing this, but I want to encourage Jou users to just `import` other Jou files by their path and not worry about build tools.

Akuli:
Fixes #823 
Fixes #806


Akuli:
Weirdly fails on LLVM 14, so must upgrade first. I'm not exactly interested in debugging it.

Akuli:
I wrote this in case the hashes ever need to be made bigger than 64 bits. It performs 16-bit multiplication without using a 16-bit data type. We could similarly perform 128-bit multiplication with a 64-bit data type.

```python3
class U16:
    low: byte
    high: byte

    def add(self: U16, other: U16) -> U16:
        result = U16{low = self.low + other.low, high = self.high + other.high}
        if result.low < self.low or result.low < other.low:
            result.high++
        return result

    def mul(self: U16, other: U16) -> U16:
        # Calculate self.low * other.low with overflow handling.
        # If other.low has first bit set, add self.low to result.
        # If other.low has second bit set, add self.low>>1 to result.
        # If other.low has third bit set, add self.low>>3 to result.
        # etc.
        result = U16{}
        shifted = U16{low = self.low, high = 0 as byte}
        n = other.low
        for i = 0; i < 8; i++:
            if n % 2 == 1:
                result = result.add(shifted)
            n /= 2 as byte

            shifted.high *= 2 as byte
            if shifted.low >= 128:
                shifted.high++
            shifted.low *= 2 as byte

        # Add 2^8 * self.low * other.high and symmetrically. Overflow doesn't
        # matter here because it would affect result by a multiple of 2^16.
        result.high += self.low * other.high
        result.high += self.high * other.low

        return result
```

Akuli:
Now calculating the checksum works fine and CI is green.

Next steps:
- [ ] Place checksum in object file names
- [ ] Do not recompile if file with correct name already exists
- [ ] Figure out some way to delete old `.o` files so that they don't accumulate over time

Akuli:
I think I will actually place two hashes to the name of each `.o` file:
- Hash of path to corresponding `.jou` file
- Hash of everything that was fed to LLVM

This way we can delete an existing compilation of the file if it cannot be reused.

Akuli:
On my computer, it now takes about one second to recompile the compiler after changing nothing:

```
akuli@akuli-desktop:~/jou$ time ./jou -o jou2 compiler/main.jou 

real	0m0,964s
user	0m0,881s
sys	0m0,075s
```

compiler/main.jou
```diff
@@ -193,6 +220,24 @@ class FileState:
 
         return mod
 
+    def compile_to_object_file(self, o_path: byte*) -> None:
```
Akuli, line 223: This looks like it should be inlined.

Akuli (commented):
- [x] read compiler/paths.jou

compiler/hash.jou
```diff
@@ -0,0 +1,118 @@
+# This file is used to calculate 64-bit hashes. They are used to check whether
+# source files have changed so that we need to recompile.
```
Akuli, line 2: Documentation folder needs to be updated

compiler/paths.jou
```diff
@@ -18,9 +20,125 @@ if WINDOWS:
...
+    #    }
+    if NETBSD:
+        const DIRENT_NAME_OFFSET: int = 13
+    else:
+        # Assume linux
+        const DIRENT_NAME_OFFSET: int = 19
+
+    if NETBSD:
+        # On NetBSD, "opendir" and "readdir" are legacy functions.
+        # We can't use them because they generate a warning at compile time.
```
Akuli, line 69:
```suggestion
        # We can't use them because they generate a linker warning.
```

compiler/paths.jou
```diff
@@ -98,97 +216,177 @@ def find_stdlib() -> byte*:
...
+                printf("Deleted %s\n", full_path)
 
-    sprintf(result, "%s/jou_compiled/%s/%s", infile_folder, subfolder, filename)
+            free(full_path)
+
+    free(list.ptr)
+
+
+def get_simple_name(jou_file_path: byte*) -> byte[50]:
+    jou_file_path = skip_slashes(jou_file_path)
```
Akuli, line 307: Todo: read rest of this file

compiler/paths.jou
```diff
@@ -98,97 +216,177 @@ def find_stdlib() -> byte*:
...
+        result[result_len++] = c
 
-    free(infile_folder)
-    free(subfolder)
     return result
 
 
+def calculate_file_path_hash(jou_file_path: byte*) -> Hash:
+    h = new_hash()
+    h.add_bytes(jou_file_path, strlen(jou_file_path))
```
Akuli, line 334: It would be better to include optimization level in this hash, so that -O0, -O1, -O2 and -O3 versions of the object file can coexist and not get deleted when recompiling with more/less optimizations.

Akuli:
- [ ] get CI green
- [ ] update documentation

PR #851, opened 2025-04-08
Use LLVM 19 for Windows builds (except --small)

Akuli:
Tested locally in a Windows 7 VM.

This is a first step towards dropping support for LLVM 14, but doesn't fully remove LLVM 14 yet. If you use `./windows_setup.sh --small`, you still get LLVM 14. I will work on that in another PR.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -102,16 +102,15 @@ jobs:
               mingw64/version_info.txt \
               $(find mingw64 -name 'crt*.o') \
               mingw64/bin/gcc.exe \
-              mingw64/lib/gcc/x86_64-w64-mingw32/12.1.0/libgcc.a \
-              mingw64/lib/gcc/x86_64-w64-mingw32/12.1.0/libgcc_eh.a \
-              mingw64/libexec/gcc/x86_64-w64-mingw32/12.1.0/liblto_plugin.dll \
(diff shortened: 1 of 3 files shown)


.github/workflows/windows.yml
```diff
@@ -102,16 +102,15 @@ jobs:
...
+              mingw64/lib/gcc/x86_64-w64-mingw32/14.2.0/libgcc.a \
+              mingw64/lib/gcc/x86_64-w64-mingw32/14.2.0/libgcc_eh.a \
+              mingw64/libexec/gcc/x86_64-w64-mingw32/14.2.0/liblto_plugin.dll \
               mingw64/x86_64-w64-mingw32/bin/ld.exe \
               mingw64/x86_64-w64-mingw32/lib/libadvapi32.a \
               mingw64/x86_64-w64-mingw32/lib/libkernel32.a \
               mingw64/x86_64-w64-mingw32/lib/libm.a \
               mingw64/x86_64-w64-mingw32/lib/libmingw32.a \
               mingw64/x86_64-w64-mingw32/lib/libmingwex.a \
-              mingw64/x86_64-w64-mingw32/lib/libmoldname.a \
```
Akuli, line 114: Don't know why, but gcc now works without this file being present. I couldn't figure out what this file even is, actually.

PR #855, opened 2025-04-08
Windows build: include mingw32-make.exe

Akuli:
Similarly to #852, this is needed to get rid of `akuli.github.io/mingw64-small.zip`, but this will make the Jou release zip files slightly bigger.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -142,7 +142,7 @@ jobs:
             mingw64/lib/libLLVMX86Info.dll.a
             mingw64/lib/libLLVMX86Desc.dll.a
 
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -141,6 +141,9 @@ jobs:
             mingw64/lib/libLLVMX86AsmParser.dll.a
             mingw64/lib/libLLVMX86Info.dll.a
             mingw64/lib/libLLVMX86Desc.dll.a
+
+            # This is here only for "./windows_setup.sh --small".
```
Akuli, line 145:
```suggestion
            # Meant to be used only for developing Jou with "./windows_setup.sh --small".
```

PR #857, opened 2025-04-08
Implement `windows_setup.sh --small` differently

Akuli:
Old implementation: Download `akuli.github.io/mingw-small.zip` instead of the full mingw64. There was a CI check that was supposed to convince users that the zip file is safe, but I don't think it was very helpful. You basically had to trust something I created by hand on my computer.

New implementation: Download a release of Jou.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,11 +30,8 @@ Following the [instructions in the README](README.md#setup) is enough.
     If you have a slow internet connection
     and it takes a long time for `windows_setup.sh` to download mingw64,
     you can instead run `./windows_setup.sh --small`.
-    This way it uses `mingw64-small.zip`,
-    which is just like the usual mingw64, but with many large files deleted to make it smaller.
-    I created it locally on my computer.
-    If you don't want to trust it, you can run `windows_setup.sh` without `--small`
-    or look at how `.github/workflows/windows.yml` compares `mingw64-small.zip` to the original `mingw64.zip`.
+    Instead of downloading the full mingw64 (about 1GB),
+    this will get a minimal version of MinGW from a [release](#releases) of Jou (about 50MB).
 5. Compile Jou:
     ```
     source activate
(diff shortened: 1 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -30,11 +30,8 @@ Following the [instructions in the README](README.md#setup) is enough.
     If you have a slow internet connection
     and it takes a long time for `windows_setup.sh` to download mingw64,
     you can instead run `./windows_setup.sh --small`.
-    This way it uses `mingw64-small.zip`,
-    which is just like the usual mingw64, but with many large files deleted to make it smaller.
-    I created it locally on my computer.
-    If you don't want to trust it, you can run `windows_setup.sh` without `--small`
-    or look at how `.github/workflows/windows.yml` compares `mingw64-small.zip` to the original `mingw64.zip`.
+    Instead of downloading the full mingw64 (about 1GB),
+    this will get a minimal version of MinGW from a [release](#releases) of Jou (about 50MB).
```
Akuli, line 34:
```suggestion
    Instead of downloading the full mingw64 (about 360MB download),
    this will get a minimal version of MinGW from a [release](#releases) of Jou (about 50MB download).
```

windows_setup.sh
```diff
@@ -48,24 +57,27 @@ else
...
     # All WinLibs versions and download links: https://winlibs.com/
     if [ $small = yes ]; then
-        # A special small version of mingw64 that comes with the repo, for people with slow internet.
-        # See .github/workflows/windows.yml for code that checks the contents of the zip so that you can trust it.
-        # TODO: this is still LLVM 14
-        url=https://akuli.github.io/mingw64-small.zip
-        filename=mingw64-small.zip
-        sha=4d858bd22f084ae362ee6a22a52c2c5b5281d996f96693984a31336873b92686
+        # User has slow internet and doesn't want to download the whole mingw64 (about 1GB).
+        # Instead, download a release of Jou (about 50MB), and extract mingw and Jou compiler from there.
```
Akuli, line 61: These are wrong and will get outdated (more wrong) easily

```suggestion
        # User has slow internet and doesn't want to download the whole mingw64.
        # Instead, download a release of Jou, and extract mingw and Jou compiler from there.
```

Issue #865, opened 2025-04-09
Comparing signed and unsigned


Akuli:
```python
import "stdlib/io.jou"

def main() -> int:
    x = -1 as int8
    y = 255 as int8
    if x == y:
        printf("wat\n")

    return 0
```

Expected result: no output
Actual result: `wat`

The problem is that `x` and `y` are compared as 8-bit values, so they are treated as equal because they consists of the same bits (11111111).

PR #866, opened 2025-04-09
Improve integer arithmetic tests

Akuli:
Now that we have more integer types (#864), we have more corner cases to test.


tests/should_succeed/compare_signed_unsigned.jou
```diff
@@ -0,0 +1,78 @@
...
+    printf("%d %d %d\n", a != bb, b != bb, c != bb)    # Output: 1 0 1
+
+    # Numbers aren't necessarily equal if they have the same bits.
+    x = -100 as int8
+    y = 156 as byte
+
+    # Output: same bits 1 1
+    printf("same bits %d %d\n", x as byte == y, y as int8 == x)
+
+    # FIXME: these tests fail
```
Akuli, line 63: I created #865

PR #867, opened 2025-04-09
Fix comparing integers of same size but different signed-ness

### LINKED ISSUE #865: Comparing signed and unsigned

Akuli:
Fixes #865

TODO:
- [ ] Uncomment the failing tests added in [#866](https://github.com/Akuli/jou/pull/866)
- [ ] Fix `int64` vs `uint64` comparing (use LLVM's 128-bit integers)
- [ ] Add tests for `int64` vs `uint64` comparing

### DIFF
--- a/compiler/typecheck/step3_function_and_method_bodies.jou
+++ b/compiler/typecheck/step3_function_and_method_bodies.jou
@@ -405,10 +405,15 @@ def check_binop(
     if got_bools:
         cast_type = boolType
     if got_integers:
-        cast_type = get_integer_type(
-            max(lhs->types.orig_type->size_in_bits, rhs->types.orig_type->size_in_bits),
-            lhs->types.orig_type->kind == TypeKind.SignedInteger or rhs->types.orig_type->kind == TypeKind.SignedInteger
-        )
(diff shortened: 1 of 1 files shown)


compiler/types.jou
```diff
@@ -351,7 +351,9 @@ def init_types() -> None:
     global_type_state.floater.type = Type{name = "float", kind = TypeKind.FloatingPoint, size_in_bits = 32 }
     global_type_state.doublelele.type = Type{name = "double", kind = TypeKind.FloatingPoint, size_in_bits = 64 }
 
-    for size = 8; size <= 64; size *= 2:
+    # Currently int8 and uint8 are not exposed in the language, but int8 is used
```
Akuli, line 354:
```suggestion
    # Currently int128 and uint128 are not exposed in the language, but int128 is used
```

Issue #350, opened 2023-03-20
What is the purpose of the 1234S syntax?


Akuli:
I'm not sure why we need the recently added `1234S` syntax. Why not just do `1234 as short`?

We still need the `L` thing to specify longs: `123456789123456789 as long` wouldn't work, because `123456789123456789` by itself would be an `int` and the value does not fit into an int.

Similarly, I think we don't really need the `12.34f` syntax. You can always do `12.34 as float`.

Akuli:
I like the idea, and it's pretty close to what Zig does, for example. But there's at least this situation where it would not work:
```
foo = 0
foo = some_function_returning_int()
```
The first line `foo = 0` sets it to be as small as possible to fit a zero, so probably 8-bit signed (currently unnamed) or `short`. But then the second line fails, because it's trying to assign `int` into a `short`.

littlewhitecloud:
```The first line foo = 0 sets it to be as small as possible to fit a zero, so probably 8-bit signed (currently unnamed) or short. But then the second line fails, because it's trying to assign int into a short.```
I mean, we should type promotion ```short``` to ```int``` not ```int``` to ```short```

**Comment by littlewhitecloud** [comment]
It will be a int, because the number doesn't fit into short(16bit), so if we don't want overflow, we have to set the type to int(32bit).
Maybe short use less memory(lol)...
It looks like python's variable (type can be change)

**Comment by Akuli** [comment]
> It looks like python's variable (type can be change)

This is something I want to avoid. When there is `foo = bar` in the code, and a variable `foo` doesn't exist yet, I want to use *only that line* to determine the type of the newly created `foo` variable. I don't want to look at other lines below to find the type of a variable. I also don't want the type of a variable to change in the middle.

Again, this is my design goal of keeping Jou simple and minimal. In the past I have made other languages where the type of a variable is determined from all places where it's used (I think my oomph language does this, for example).

**Comment by littlewhitecloud** [comment]
I mean we can change the exist variable type, not create a new one.

**Comment by Akuli** [comment]
Another thought: Changing the type of a variable in the middle becomes complicated when the same code executes many times.
```
foo = 0  # foo is int
while True:
    # foo is int
    printf("%d\n", foo)
    ...
    foo = 123456789123456789L
    # foo is long
```

If the loop runs more than once, we somehow have to convert `123456789123456789L` to `int` because the type of the variable was changed. The simplest solution is to never change the type of an existing variable.

Akuli:
[...]

PR #871, opened 2025-04-16
Infer types of integers in code

### LINKED ISSUE #863: Large integer literals
https://github.com/Akuli/jou/blob/266f525ffc5237b59ea56ca99cb5d667bd6528d2/compiler/hash.jou#L68-L82

It is currently not possible to express a 64-bit unsigned value that doesn't fit in a 64-bit signed integer. This would be the case even if there was a 64-bit unsigned integer type (#164).

I think initializing integer should work similarly to string literals: `"foo"` is usually `byte*`, but can also be e.g. `byte[4]` or `byte[100]` if a type annotation or other hint is given. Similarly, `123` could usually be int, but if a type hint is given, it could also be e.g. `byte` or `long` or `unsigned long`:

```python
a = 123  # type defaults to int
b = 123123123123123  # Error: value is too large for int (32-bit signed integer)
c: long = 123123123123123
d: byte = 123
e: byte = 1234  # Error: value is too large for byte (8-bit unsigned integer)
f: short = 1234
```

Advantages:
- Less syntax: no more `123S` or `123L` or other suffixes to memorize
- Easy to implement, because strings are similar

Disadvantages:
- Seems slightly more magical to users

Akuli:
If you have a function that takes an `int64` (aka `long`) as an argument, you can now do:

```python
my_function(123123123123123)
```

Previously this didn't work because the type of `123123123123123` was `int`, not `int64`, and `123123123123123` does not fit into 32 bits. You had to do `my_function(123123123123123L)`, where the `L` at the end stands for `long`. IMO this is unnecessary and surprising.

While the new behavior may feel a bit magical at first, it is still much simpler than many other languages, and fully documented.

Fixes #863. Discussed in #350.


compiler/evaluate.jou
```diff
@@ -39,11 +39,22 @@ def evaluate_boolean_and_or(expr: AstExpression*, and_or: byte*) -> int:
 
 
 # Returns False if the expression is not a simple compile-time constant.
+#
+# If a type hint is given, does not necessarily return a value of that type.
+# As the name suggests, it is merely a hint.
```
Akuli, line 44:
```suggestion
```

compiler/evaluate.jou
```diff
@@ -39,11 +39,22 @@ def evaluate_boolean_and_or(expr: AstExpression*, and_or: byte*) -> int:
...
 @public
 def evaluate_constant_expression(expr: AstExpression*, result: Constant*) -> bool:
     msg: byte[500]
 
     match expr->kind:
+        case AstExpressionKind.IntegerConstant:
+            # TODO: don't assume "int" when it doesn't make sense
+            if expr->integer_value <= 0x7fff_ffff:
```
Akuli, line 52:
```suggestion
            if (expr->integer_value as int) == expr->integer_value:
```

compiler/token.jou
```diff
@@ -9,9 +9,9 @@ declare isprint(b: int) -> int
 
 @public
 enum TokenKind:
-    Short
-    Int
-    Long
+    Short  # old, bad, basically deprecated, e.g. 123S
+    Long   # old, bad, basically deprecated, e.g. 123123123123123L
```
Akuli, line 13: I feel this way, but let's leave it for another PR to keep this self-contained

```suggestion
    Short  # e.g. 123S
    Long   # e.g. 123123123123123L
```

compiler/tokenizer.jou
```diff
@@ -72,37 +69,46 @@ def parse_integer(string: byte*, location: Location, nbits: int) -> long:
...
 
     if nbits == 16 and (result as short) != result:
         overflow = True
 
     if overflow:
-        message = malloc(100)
-        sprintf(message, "value does not fit in a signed %d-bit integer", nbits)
-        fail(location, message)
+        if nbits == 0:
+            # TODO: test this, if not already tested
```
Akuli, line 106: There is test

```suggestion
```

compiler/tokenizer.jou
```diff
@@ -72,37 +69,46 @@ def parse_integer(string: byte*, location: Location, nbits: int) -> long:
...
         # Overflow isn't UB in Jou
-        if (result * base) / base != result:
+        if (result * (base as uint64)) / (base as uint64) != result:
             overflow = True
             break
-        result *= base
+        result *= base as uint64
 
-        if result + hexdigit_value(digits[i]) < 0:
+        if result + digit_value < result:
```
Akuli, line 91: Not sure whether this is working, let's make it more robust

```suggestion
        if (result + digit_value) as uint64 < result:
```

compiler/typecheck/step3_function_and_method_bodies.jou
```diff
@@ -775,13 +777,66 @@ def handle_conflicting_class_field_and_enum_member_syntax(state: State*, expr: A
...
+
 def typecheck_expression(state: State*, expr: AstExpression*, type_hint: Type*) -> None:
     msg: byte[500]
     result: Type* = NULL
 
     handle_conflicting_class_field_and_enum_member_syntax(state, expr)
 
     match expr->kind:
+        case AstExpressionKind.IntegerConstant:
+            # Usually int, but use type hint if given resaonably.
```
Akuli, line 822:
```suggestion
            # Usually int, but use type hint if given reasonably.
```

tests/should_succeed/integer_type_inference.jou
```diff
@@ -0,0 +1,35 @@
+import "stdlib/io.jou"
+
+
+def print_byte(b: byte) -> None:
+    printf("%c\n", b)
+
+
+def main() -> int:
+    # Type of 250 is inferred to be byte
```
Akuli, line 9:
```suggestion
    # Type of 100 is inferred to be byte. It never becomes int, then assigning to byte variable would fail.
```

Issue #515, opened 2024-08-19
Rename long to int64?


Akuli:
C has `long` that is 32 bits on Windows and 64 bits on many other platforms. Jou has `long` which is always 64 bits. This causes confusion when combining C code with Jou code. That's bad, because working with C is a very important feature in Jou.

I think Jou `long` should be renamed to `int64`. @littlewhitecloud What do you think?

littlewhitecloud:
> C has that is 32 bits on Windows and 64 bits on many other platforms. Jou has which is always 64 bits. This causes confusion when combining C code with Jou code. That's bad, because working with C is a very important feature in Jou.`long``long`
> 
> I think Jou should be renamed to . @littlewhitecloud What do you think?`long``int64`

It's up to you :)
Other programming language seems use names like this too (`i64` `i32`) (`rust`?)

Akuli:
I don't want Jou to use `i32` because Python programmers know `int`, and `int` is almost always 32 bits nowadays.

Akuli:
I think integer types should be:

| bits | signed | unsigned |
|---|---|---|
| 8 | `int8` | `byte` |
| 16 | `int16` | `uint16` |
| 32 | `int` | `uint` |
| 64 | `int64` | `uint64` |

Advantages:
- Simple
- Relatively short
- Somewhat consistent with other languages
- Easy to tell how many bits each data type is

Disadvantages:
- `int8` and `byte` are inconsistent.

I don't want to use `signed byte` because then `signed byte` is the only type with a two-word name.

I don't want to use `uint8` because then every Jou beginner would ask "is that a byte?".

Akuli:
I'm no longer convinced that this is a good idea. Many other languages have a 64-bit `long`, but I don't know any language that mixes names with numbers (`int64`) and without numbers (`int`).

Akuli:
There is now `int64` type (#864). I think I will remove `long`.

PR #875, opened 2025-04-18
Delete the `short` data type

Akuli:
Use `int16` instead of `short`. It does the same thing but the name is more explicit.

Also, `short` was never really used anywhere. The main use case I can think of is audio: a typical `.wav` file is a few headers and a big list of 16-bit integers. I haven't done much audio stuff with Jou.

Related to #515

### DIFF
--- a/tests/should_succeed/printf.jou
+++ b/tests/should_succeed/printf.jou
@@ -3,9 +3,9 @@ import "stdlib/str.jou"
 
 def main() -> int:
     printf("Hello world\n")  # Output: Hello world
-    printf("short %hd\n", 11451 as short) # Output: short 11451
-    printf("number %d\n", 123)  # Output: number 123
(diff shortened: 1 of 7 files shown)


tests/should_succeed/printf.jou
```diff
@@ -3,9 +3,9 @@ import "stdlib/str.jou"
 
 def main() -> int:
     printf("Hello world\n")  # Output: Hello world
-    printf("short %hd\n", 11451 as short) # Output: short 11451
-    printf("number %d\n", 123)  # Output: number 123
-    printf("long %lld\n", 123451234512345 as long)  # Output: long 123451234512345
+    printf("16-bit %d\n", 11451 as int16) # Output: 16-bit 11451
+    printf("32-bit %d\n", 123)  # Output: number 123
```
Akuli, line 7:
```suggestion
    printf("32-bit %d\n", 123)  # Output: 32-bit 123
```

Issue #879, opened 2025-04-20
Bitwise operators


Akuli:
I profiled the compiler with `valgrind --tool=callgrind` and `kcachegrind`. When it is not invoking LLVM or the linker, most of the time is spent calculating hashes of the code that would be fed into LLVM. It boils down to this function:

https://github.com/Akuli/jou/blob/4ed93c98a98bcd6290ececa3109d26d9045891d5/compiler/hash.jou#L21-L33

To speed up this (and other similar things), Jou should have bitwise operators, just like C:
- [x] and `&`
- [x] or `|`
- [x] not `~`
- [x] xor `^`
- [ ] shift left `<<`
- [x] shift right `>>`

I will probably start with xor because I have limited time to spend on Jou and xor would be the most useful to have.

PR #883, opened 2025-04-20
Add bitwise AND (`&`) and OR (`|`) operators

Akuli:
Checks two boxes in #879 

Examples:

```python
import "stdlib/io.jou"

def main() -> int:
    # Add the bit that makes letters lowercase
    printf("%c\n", 'A' | 0x20)  # Output: a

    # Remove the bit that makes letters lowercase
    printf("%c\n", 'a' & 0xDF)  # Output: A

    return 0
```


compiler/parser.jou
```diff
@@ -756,7 +767,7 @@ class Parser:
     # TODO: ban chaining different bitwise operations e.g. a ^ b << c
```
Akuli, line 767:
```suggestion
    # TODO: ban chaining different bitwise operations e.g. a ^ b & c
```

examples/aoc2024/day17/part2.jou
```diff
@@ -28,14 +28,14 @@ def get_first_output(code: int*, reg_a: long) -> int:
...
-            regs[1] = do_combo_op(operand, regs) % 8
+            regs[1] = do_combo_op(operand, regs) & 0b111
         elif opcode == 3:  # jnz = Jump if NonZero
             if regs[0] != 0:
                 ip = &code[operand]
         elif opcode == 4:  # bxc = B Xor C
             regs[1] ^= regs[2]
         elif opcode == 5:  # out = OUTput value
-            return (do_combo_op(operand, regs) % 8) as int
+            return (do_combo_op(operand, regs) & 0b111) as int
```
Akuli, line 38: I think `as int` is unnecessary due to Jou's rules for determining the type of `&`. It is the smaller of the two types (or the unsigned one if the same size).

examples/aoc2024/day17/part2.jou
```diff
@@ -50,23 +50,16 @@ def get_first_output(code: int*, reg_a: long) -> int:
...
-    while how_many --> 0:
-        power_of_two *= 2
-    return n ^ ((n / power_of_two) * power_of_two)
-
-
 # My input looked like this, as printed by code in part 1:
 #
 #    Registers: A=... B=0 C=0
-#       0  B = A % 8
+#       0  B = A & 0b111
```
Akuli, line 56: Let's not change the comment
```suggestion
#       0  B = A % 8
```

examples/aoc2024/day17/part2.jou
```diff
@@ -50,23 +50,16 @@ def get_first_output(code: int*, reg_a: long) -> int:
...
 #       2  B ^= 1
 #       4  C = A / 2**B
 #       6  B ^= 5
 #       8  B ^= C
 #      10  A /= 2**3
-#      12  output B % 8
+#      12  output B & 0b111
```
Akuli, line 62: Let's not change the comment
```suggestion
#      12  output B % 8
```

examples/aoc2024/day17/part2.jou
Akuli:
```suggestion
            return do_combo_op(operand, regs) & 0b111
```

PR #885, opened 2025-04-21
Improve markdown link checker

Akuli:
- Do not complain if a link has been down for less than 3 days.
- Delete the retrying logic. It is no longer needed.
- Use a hard-coded/pinned version of Python's `requests` library. This is better for security (although `requests` is very unlikely to get hacked) and makes the check less likely to break by itself.

### DIFF
--- a/check-markdown-links.py
+++ b/check-markdown-links.py
@@ -2,6 +2,7 @@
 # Get from https://github.com/Akuli/porcupine/
 
 import argparse
+import json
 import os
 import re
 import subprocess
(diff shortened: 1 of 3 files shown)


check-markdown-links.py
```diff
@@ -2,14 +2,14 @@
 # Get from https://github.com/Akuli/porcupine/
```
Akuli, line 2:
```suggestion
# There is a similar script in https://github.com/Akuli/porcupine/
```

PR #886, opened 2025-04-21
`const`: Add support for negative values and all integer types

Akuli:
Works now, didn't work before:

```python
const big_number: int64 = 12345678987654321
```


compiler/evaluate.jou
```diff
@@ -38,19 +38,88 @@ def evaluate_boolean_and_or(expr: AstExpression*, and_or: byte*) -> int:
...
-                *result = int_constant(intType, expr->integer_value as int64)
-                return True
-            return False
+            if type_hint != NULL and type_hint->is_integer_type():
+                t = type_hint
+            else:
+                t = intType
+
+            if expr->integer_value > max_value(t):
+                # TODO: test this in a constant
```
Akuli, line 113: must do

compiler/evaluate.jou
```diff
@@ -93,15 +162,18 @@ def evaluate_constant_expression(expr: AstExpression*, result: Constant*) -> boo
...
+            if not evaluate_constant_expression(&expr->operands[0], result, type_hint):
+                return False
+
+            if result->get_type()->kind != TypeKind.SignedInteger:
                 return False
 
-            # TODO: make this more general
-            if result->get_type() != intType:
+            # Negating the smallest signed integer would overflow.
+            # For example, int8 goes from -128 to +127.
```
Akuli, line 172:
```suggestion
            # For example, int8 goes from -128 to +127.
            # TODO: test this when possible
```

compiler/typecheck/step3_function_and_method_bodies.jou
```diff
@@ -878,23 +847,12 @@ def typecheck_expression(state: State*, expr: AstExpression*, type_hint: Type*)
...
-                if result->kind == TypeKind.SignedInteger:
-                    signed_or_unsigned = "signed"
-                else:
-                    signed_or_unsigned = "unsigned"
-                snprintf(msg, sizeof(msg), "value does not fit into %s (%d-bit %s integer)", result->name, result->size_in_bits, signed_or_unsigned)
-                fail(expr->location, msg)
-
+            c: Constant
+            ok = evaluate_constant_expression(expr, &c, type_hint)
+            assert ok  # See source code of evaluate_constant_expression(), should never return False
```
Akuli, line 852:
```suggestion
            assert ok  # Should fail with a compiler error if there is a problem
```

PR #887, opened 2025-04-21
Compiler code cleanup: use bitflags for UVG statuses

Akuli:
Seems like a slightly nicer way to do the thing.

### DIFF
--- a/compiler/uvg_analyze.jou
+++ b/compiler/uvg_analyze.jou
@@ -38,6 +38,8 @@ if False:
                     printf("{u,d}")
                 case DONT_ANALYZE:
                     printf("-")
+                case _:
+                    printf("?")
         printf("\n")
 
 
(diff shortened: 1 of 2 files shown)


compiler/uvg_analyze.jou
```diff
@@ -8,120 +8,71 @@ import "./uvg.jou"
...
                     printf("{u}")
-                case VarStatus.Defined:
+                case DEFINED:
                     printf("{d}")
-                case VarStatus.PossiblyUndefined:
+                case POSSIBLY_UNDEFINED:
                     printf("{u,d}")
-                case VarStatus.DontAnalyze:
+                case DONT_ANALYZE:
                     printf("-")
```
Akuli, line 40:
```suggestion
                    printf("-")
                case _:
                    printf("?")
```

PR #888, opened 2025-04-21
Fix a few typos

Akuli:
Actually, add codespell to CI so that it will complain if there are typos. (The title of this issue is written so that it shows up nicely in a changelog.)

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -17,9 +17,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - run: which codespell || true  # TODO: remove this
-      - run: sudo apt update && sudo apt install codespell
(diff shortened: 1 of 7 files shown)


.github/workflows/checks.yml
```diff
@@ -12,6 +12,15 @@ jobs:
       - uses: actions/checkout@v4
       - run: shellcheck --color=always --shell=bash --exclude=SC2086,SC2059,SC2046,SC2235,SC2002,SC2206,SC2068,SC2207,SC2013 *.sh activate
 
+  codespell:
+    timeout-minutes: 2
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: sudo apt update
+      - run: sudo apt install codespell
```
Akuli, line 21:
```suggestion
      - run: sudo apt install -y codespell
```

PR #889, opened 2025-04-28
Add compile-time asserts

### LINKED ISSUE #882: Compile-time assert
Asserts outside functions should perhaps be evaluated at compile time, similarly to compile-time if statements.

Simple example:

```python
const a: int = 1
assert a == 1
```

Practical example:

```python
# this file should never be imported on windows
assert not WINDOWS

def foo() -> None:
    ...posix stuff...
```

Akuli:
It is now possible to write `assert not WINDOWS` outside a function. This way, you can avoid accidentally importing a Windows-specific module on Linux, for example.

Before it produced the following error:

```
compiler error in file "a.jou", line 1: 'assert' must be inside a function or method
```

This is similar to `static_assert` in C/C++.

Fixes #882.

### DIFF
--- a/compiler/typecheck/step3_function_and_method_bodies.jou
+++ b/compiler/typecheck/step3_function_and_method_bodies.jou
@@ -1460,5 +1460,17 @@ def typecheck_step3_function_and_method_bodies(ast: AstFile*) -> None:
                 for inner = stmt->classdef.body.ptr; inner < stmt->classdef.body.end(); inner++:
                     if inner->kind == AstStatementKind.MethodDef:
                         typecheck_function_or_method_body(&state, &inner->method)
+            case AstStatementKind.Assert:
+                c: Constant
+                if not evaluate_constant_expression(&stmt->assertion.condition, &c, boolType):
+                    # TODO: test this
+                    fail(stmt->location, "cannot evaluate assertion at compile time")
+                if c.get_type() != boolType:
+                    # TODO: test this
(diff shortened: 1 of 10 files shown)


compiler/typecheck/step3_function_and_method_bodies.jou
```diff
@@ -1460,5 +1460,17 @@ def typecheck_step3_function_and_method_bodies(ast: AstFile*) -> None:
                 for inner = stmt->classdef.body.ptr; inner < stmt->classdef.body.end(); inner++:
                     if inner->kind == AstStatementKind.MethodDef:
                         typecheck_function_or_method_body(&state, &inner->method)
+            case AstStatementKind.Assert:
+                c: Constant
+                if not evaluate_constant_expression(&stmt->assertion.condition, &c, boolType):
+                    # TODO: test this
```
Akuli, line 1466:
```suggestion
```

compiler/typecheck/step3_function_and_method_bodies.jou
```diff
@@ -1460,5 +1460,17 @@ def typecheck_step3_function_and_method_bodies(ast: AstFile*) -> None:
...
+                    fail(stmt->location, "cannot evaluate assertion at compile time")
+                if c.get_type() != boolType:
+                    # TODO: test this
```
Akuli, line 1469:
```suggestion
```

PR #896, opened 2025-06-20
Add --fail-on-warnings command line argument

### LINKED ISSUE #892: "No warnings allowed" compiling mode
Compiler warnings are often acceptable during development, but should be fixed before the code ends up in the main branch.

The Jou compiler should have a mode, perhaps enabled by a command line flag, where it exits with error code 1 (failure) if there are any compiler warnings. This mode would be off by default, but it would be useful to enable it when compiling Jou code in GitHub Actions and other CI.

In other words, this would be just like `-Werror` in C compilers.

Akuli:
Fixes #892

### DIFF
--- a/tests/should_succeed/compiler_cli.jou
+++ b/tests/should_succeed/compiler_cli.jou
@@ -73,11 +73,18 @@ def main() -> int:
 
     # Test treating compiler warnings as errors
     # Output: compiler warning for file "unused_enum.jou", line 3: enum Foo defined but not used
-    # Output: compiler error: using exit code 1 because --fail-on-warnings was given
+    # Output: compiler error: there were warnings and --fail-on-warnings was given
+    # Output: compiler warning for file "unused_enum.jou", line 3: enum Foo defined but not used
+    # Output: compiler error: there were warnings and --fail-on-warnings was given
     if WINDOWS:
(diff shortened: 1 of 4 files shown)


tests/should_succeed/compiler_cli.jou
```diff
@@ -71,6 +71,22 @@ def main() -> int:
...
     print_reusing_compiled_file_lines("tmp/hello_out_v.txt")
 
+    # Test treating compiler warnings as errors
+    # Output: compiler warning for file "unused_enum.jou", line 3: enum Foo defined but not used
+    # Output: compiler error: there were warnings and --fail-on-warnings was given
+    # Output: compiler warning for file "unused_enum.jou", line 3: enum Foo defined but not used
+    # Output: compiler error: there were warnings and --fail-on-warnings was given
+    if WINDOWS:
+        system("cd tests\\should_succeed & ..\\..\\jou.exe --fail-on-warnings unused_enum.jou && echo This should not be shown")
+        system("cd tests\\should_succeed & ..\\..\\jou.exe -o foo --fail-on-warnings unused_enum.jou && echo This should not be shown")
```
Akuli, line 81:
```suggestion
        system("cd tests\\should_succeed && ..\\..\\jou.exe --fail-on-warnings unused_enum.jou && echo This should not be shown")
        system("cd tests\\should_succeed && ..\\..\\jou.exe -o foo --fail-on-warnings unused_enum.jou && echo This should not be shown")
```

Issue #749, opened 2025-02-07
The `.` vs `->` issue


Akuli:
(This has been discussed in several other issues in the past.)

Currently Jou has two operators for accessing attributes:
- `foo.bar`: Used when `foo` is an instance of a class. This means "take the instance `foo` and access its `bar` attribute".
- `foo->bar`: Used when `foo` is a pointer to an instance. This means "take the instance at the memory location `foo` and access its `bar` attribute". So it's same as `(*foo).bar`.

The problem is that @Moosems and @littlewhitecloud don't like `->`, and they would like to use `.` for everything.

Advantages of `.` for everything:
- Looks pretty
- Feels consistent: always the same operator, no matter what you are doing

Disadvantages of using `.` for everything:
- Hard to tell whether we are accessing the memory where a pointer points, which may crash if the pointer is bad
- Inconsistent with C

Let's look at how other languages do it:

| Language | Without pointer | With pointer | Notes |
|----|----|----|---|
| C | `.` | `->` ||
| C++ | `.` | `->` or `.` | Same as C, but also has references which use `.` |
| Python | (missing) | `.` | Objects are always accessed through pointers |
| Java | (missing) | `.` | Objects are always accessed through pointers |
| JavaScript | (missing) | `.` | Objects are always accessed through pointers |
| Rust | `.` | `.` | Also has `->` but it's rarely used |
| Zig | `.` | `.` |  |
| Jou as it is now | `.` | `->` | Same as C |
| Jou as proposed | `.` | `.` | Same as Rust and Zig |

Moosems:
I think there's a way that we can appease both halves of this. A pointer deref prefix vs postfix. If I use `&` as my prefix, for example, `&foo.bar` would be fine to me. The issue is `->` handling both accessing and dereferencing

Akuli:
`*` is the dereference operator in basically all languages that have pointers, so let's use that :)

> If I use & as my prefix, for example, &foo.bar would be fine to me.

The problem is that `*foo.bar` already does a different thing. It is equivalent to `*(foo.bar)`, so it interprets `foo` as an instance of class and its field `bar` as a pointer. This is often quite useful. We cannot change the meaning of `*foo.bar` because then it would be different from C, and it would cause a lot of pain to anyone trying to port C code to Jou, or vice versa. C compatibility is important.

---

> A pointer deref prefix vs postfix.

[...]

PR #899, opened 2025-07-28
Allow using `.` instead of `->` to look up members from pointers

Akuli:
Example:

```python
import "stdlib/io.jou"

class Foo:
    x: int

def move(f: Foo*) -> None:
    #f->x++  # old syntax
    f.x++
```

Before this PR, there were two ways to access attributes: `foo.bar` and `foo->bar`. You used `foo.bar` if `foo` is an instance and `foo->bar` if `foo` is a pointer to an instance.

This PR modifies `.` to work for both use cases. For backwards compatibility, `foo->bar` will do the same thing as `foo.bar` for a few hours. Then I will remove the `foo->bar` syntax.

Todo:
- [x] update docs

Part of #749

### DIFF
--- a/compiler/builders/ast_to_builder.jou
+++ b/compiler/builders/ast_to_builder.jou
@@ -417,7 +417,7 @@ class AstToBuilder:
                 # We can't do this for all expressions, because e.g. &some_function() is error.
                 return self->builder->dereference(self->build_address_of_expression(expr))
             case AstExpressionKind.DotOperator:
-                # Compiler should replace these with other expression kinds when it knows what they are
+                # Should become other expression kind before we get here
                 assert False
             case AstExpressionKind.GetEnumMember:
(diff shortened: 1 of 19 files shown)


compiler/builders/ast_to_builder.jou
```diff
@@ -416,18 +416,22 @@ class AstToBuilder:
                 # ptr[foo] can always be evaluated as &ptr[foo], because ptr is already a pointer.
                 # We can't do this for all expressions, because e.g. &some_function() is error.
                 return self->builder->dereference(self->build_address_of_expression(expr))
+            case AstExpressionKind.DotOperator:
+                # Compiler should replace these with other expression kinds when it knows what they are
```
Akuli, line 420:
```suggestion
                # Should become other expression kind before we get here
```

PR #902, opened 2025-07-28
Use `.` instead of `->` everywhere

Akuli:
Huge PR. Generated by applying `sed -E -i 's/([^ ])->([^ ])/\1.\2/g'` to all files in the repository.

Part of #749


Akuli:
I used a script to do this and it replaced a bit too much. Let me do it differently.

PR #909, opened 2025-08-02
Improve stdlib/math.jou

Akuli:
- Add `M_PI` (3.1415...) and `M_E` (2.71828...) constants.
- Delete functions that don't exist on NetBSD: `exp10`, `exp10f`
- Delete functions that nobody uses: `erfc`, `erfcf`, `tgamma`, `tgammaf`, `lgamma`, `lgammaf`
- Improve documentation comments.
- Rearrange functions (e.g. `sin` before `cos`) to make the file more readable.

### DIFF
--- a/stdlib/math.jou
+++ b/stdlib/math.jou
@@ -193,12 +193,6 @@ declare exp(x: double) -> double
 @public
 declare expf(x: float) -> float
 
-# Computes 10^x (10 to the power of x).
-@public
-declare exp10(x: double) -> double
-@public
-declare exp10f(x: float) -> float
-
 # Computes x^y (x to the power of y).
 @public
 declare pow(x: double, y: double) -> double
--- a/stdlib/math.jou
+++ b/stdlib/math.jou
@@ -240,3 +240,28 @@ declare cbrtf(x: float) -> float
 declare hypot(x: double, y: double) -> double
 @public
 declare hypotf(x: float, y: float) -> float
+
+# Computes the error function, defined by the following integral:
+#
+#                           x
+#                          |\
+#   erf(x) = 2/sqrt(pi) *  |  exp(-t^2) dt
+#                         \|
+#                          0
+#
+# The scaling factor in front is chosen such that:
+#
+#   erf(infinity) == 1
+#   erf(-infinity) == -1
+#
+# This can be used to calculate probabilities that follow the Gaussian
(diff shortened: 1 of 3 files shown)


stdlib/math.jou
```diff
@@ -111,110 +155,88 @@ declare nan(tagp: byte*) -> double
...
 @public
 declare hypot(x: double, y: double) -> double
 @public
 declare hypotf(x: float, y: float) -> float
-
-# Error and gamma functions
-
-# Returns the error function value for x.
-@public
-declare erf(x: double) -> double
```
Akuli, line 203: I think I should add back `erf` and `erff`. Without these, there's no good way to compute the integral of `exp(-x^2)`, which is needed for many practical probability calculations.

Basically anything that is a sum or average of many similar things is gaussian distributed. And to work out the probability of the thing being less than or greater than a given value, you need these functions.

stdlib/math.jou
```diff
@@ -48,56 +68,82 @@ declare fabsf(x: float) -> float
...
 declare asinf(x: float) -> float
+
+# Returns an angle between 0 and pi so that `cos(angle) == x`.
+# x should be between -1 and 1.
+@public
+declare acos(x: double) -> double
+@public
+declare acosf(x: float) -> float
+
+# Returns an angle between -pi/2 and pi/2 so that `tan(angle) == x`.
```
Akuli, line 138:
```suggestion
# Returns an angle between -pi/2 and pi/2 so that `tan(angle) == x`.
# `x` can be any number.
```

stdlib/math.jou
```diff
@@ -48,56 +68,82 @@ declare fabsf(x: float) -> float
...
 # A full turn (360 degrees) is 2pi (about 6.28) radians.
-@public
-declare cos(x: double) -> double
-@public
-declare cosf(x: float) -> float
+#
+# Use the following formulas to convert between radians and degrees:
+#
+#   radians = degrees / 360.0 * (2*M_PI)
+#   degrees = radians / (2*M_PI) * 360
```
Akuli, line 103: Later I might add functions to convert between these two like Python's `math.degrees` and `math.radians`. But I will name them less confusingly: `math.radians(90)` is correct to convert 90 degress to radians, but it looks a lot like 90 radians and that's not what it is at all.

PR #917, opened 2025-08-05
Create stdlib/sort.jou

### LINKED ISSUE #915: Add qsort to standard library
Now that we have function pointers, it is possible to call `qsort()` from Jou code to sort an array. Here's how it is done:

```python
import "stdlib/io.jou"

declare qsort(
    base: void*,
    nmemb: int64,
    size: int64,
    compare: funcptr(void*, void*) -> int,
) -> None

def compare_ints(a: void*, b: void*) -> int:
    return *(a as int*) - *(b as int*)

def main() -> int:
    array = [5, 1, 4, 3, 6, 2]
    qsort(array, 6, sizeof(array[0]), compare_ints)

    # Output: 1 2 3 4 5 6
    for i = 0; i < 6; i++:
        if i != 0:
            printf(" ")
        printf("%d", array[i])
    printf("\n")

    return 0
```

I should add `qsort()` to standard library.

Akuli:
This PR creates `stdlib/sort.jou`. It is based on the C `qsort()` function.

Unlike `qsort()`, Jou's sorting is typesafe. Practically this means that you don't need `void*` pointers or two different `sizeof()` calls to sort a list, so it is much harder to mess it up. This is implemented with a generic `class Sorter[T]` whose only purpose is to provide a method `sort()` that calls `qsort()` in the right way for sorting elements of type `T`.

For convenience, there are also functions to sort lists of integers or strings.

There is also tests and documentation.

Fixes #915


stdlib/sort.jou
```diff
@@ -0,0 +1,81 @@
...
+#
+# See doc/sorting.md for documentation.
+
+import "stdlib/str.jou"
+
+declare qsort(
+    base: void*,
+    nmemb: int64,
+    size: int64,
+    compare: funcptr(void*, void*) -> int,
```
Akuli, line 11: Should rename these parameters so they are less cryptic

stdlib/sort.jou
```diff
@@ -0,0 +1,81 @@
...
+# integer. The return value should be:
+#
+#   negative    if first < second
+#   0           if first == second
+#   positive    if first > second
+@public
+class Sorter[T]:
+    @inline
+    def sort(self: Sorter[T], ptr: T*, len: int64, compare: funcptr(T*, T*) -> int) -> None:
+        qsort(ptr, len, sizeof(ptr[0]), compare as funcptr(void*, void*) -> int)
```
Akuli, line 28: Should probably rename `compare` to `comparator`, because that's what docs call it.

tests/should_succeed/sort_test.jou
```diff
@@ -0,0 +1,92 @@
...
+
+    def print(self) -> None:
+        printf("%s (%d)\n", self.name, self.year)
+
+
+def compare_language_name(a: Language*, b: Language*) -> int:
+    return strcmp(a.name, b.name)
+
+def compare_language_year(a: Language*, b: Language*) -> int:
+    return ((a.year > b.year) as int) - ((a.year < b.year) as int)
```
Akuli, line 18: Let's use the documented way:

```suggestion
    return a.year - b.year
```

PR #930, opened 2025-08-10
Temp test

Akuli:
Do not merge. Checking if a broken unmerged PR that changes `macos.yml` creates a new entry in #928.


Akuli:
Yes, it does.

PR #933, opened 2025-08-10
Clarify how float and double literals work

Akuli:
I noticed that these were defined somewhat ambiguously, so I rewrote the definitions in a way that makes sense and changed the compiler accordingly.

### DIFF
--- a/compiler/tokenizer.jou
+++ b/compiler/tokenizer.jou
@@ -103,7 +103,7 @@ def is_valid_double(str: byte*) -> bool:
     # See doc/syntax-spec.md
     e = strstr(str, "e")
     if e == NULL:
-        # Zero or more digits, then '.', then zero or more digits, but cannot be just "."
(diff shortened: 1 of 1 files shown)


compiler/tokenizer.jou
```diff
@@ -98,27 +98,32 @@ def parse_integer(string: byte*, location: Location) -> uint64:
...
-        # 123.456
-        e == NULL and minus == NULL and dot != NULL
-    ) or (
-        # 1e-4
-        e != NULL and e[1] != '\0' and (dot == NULL or dot < e) and  (minus == NULL or (&e[1] == minus and e[2] != '\0'))
-    )
+    # See doc/syntax-spec.md
+    e = strstr(str, "e")
+    if e == NULL:
+        # Zero or more digits, then '.', then zero or more digits, but cannot be just "."
```
Akuli, line 106:
```suggestion
        # One or more digits, then '.', then one or more digits
```

PR #936, opened 2025-08-21
modify perlin noise example image file format to ppm

olkku45:
there is the first commit as well, which is the completely faulty version :P


Akuli:
As it is now, this PR does a lot more than changing the output image format:
- Size is now specified with four constants instead of two, and they can get out of sync
- The code has been rewritten quite a lot

Ideally, this PR would only change the output image format.

All you really need to do is to replace `stbi_image_write()` with a new function. The existing code already constructed the contents that go after the header (`image` variable), so you don't need to change how that's done. Just write the whole thing to file with one `fwrite()`.

Can you do that?

olkku45:
Okay. Now it's way simpler and doesn't rewrite anything. Also, consider adding fwrite() since it doesn't seem to be in stdlib/io!

Akuli (approved):
Thanks for doing this! For some reason, I didn't notice that you had pushed a new commit.

PR #940, opened 2025-09-06
Recompile when type of local variable changes

### LINKED ISSUE #939: Array size cache bug
Consider the following:

```python
import "stdlib/io.jou"

def main() -> int:
    foo: byte[10]

    for i = 0; i < 100; i++:
        foo[i] = 'x'
    foo[100] = 0

    puts(foo)
    return 0
```

This is broken, because the `foo` array is too small. It crashes when I run the program.

The problem is that the crash does not go away if I change `foo: byte[10]` to `foo: byte[1000]`. The program is not recompiled when I run `jou file.jou`. But if I delete `jou_compiled`, it will start working:

```
akuli@akuli-desktop:~/jou$ jou bug.jou
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Segmentation fault
akuli@akuli-desktop:~/jou$ nano bug.jou   # change 10 to 1000
akuli@akuli-desktop:~/jou$ jou bug.jou 
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Segmentation fault
akuli@akuli-desktop:~/jou$ rm -rf jou_compiled/
akuli@akuli-desktop:~/jou$ jou bug.jou 
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
akuli@akuli-desktop:~/jou$ 
```

(I noticed this when explaining a thing to @olkku45.)

Akuli:
Fixes #939

### DIFF
--- a/compiler/builders/hash_builder.jou
+++ b/compiler/builders/hash_builder.jou
@@ -109,6 +109,7 @@ class HBuilder:
         # Do nothing with varname, so that renaming a local variable doesn't cause a rebuild
         # TODO: Will need to be different if debug info is enabled
         self.hash.add_string("stack_alloc")
+        hash_type(&self.hash, t)
         return self.new_value(t.pointer_type())
 
     # *ptr = value


Akuli:
CI failure is because llvm.org is down right now.

Issue #561, opened 2025-01-08
Jou on embedded


Akuli:
From DMs with @moosems:

![screenshot-1736346404](https://github.com/user-attachments/assets/7828740a-deb9-446e-8933-e950995ee27a)

This will probably require:
- changes to compiler (supporting different architectures, in LLVM and when linking)
- changes to the language (`if WINDOWS:` doesn't really make sense on embedded)
- documentation

Moosems:
A special embedded api for pinout would likely be necessary as well

Moosems:
How does target compilation work right now?

Akuli:
Very much hard-coded / always using whatever system the compiler is running on:

https://github.com/Akuli/jou/blob/2b33658b17239c68c84b034c33c2a7c145da43dd/compiler/target.jou#L27-L49

Akuli:
Another problem is that 64-bitness is hard-coded into various places. C's `size_t` type is 32-bit or 64-bit depending on the platform. Jou assumes it is 64-bit.

For example, consider the `strlen()` function:

![Image](https://github.com/user-attachments/assets/57a22909-4002-467f-886e-d777dedb2e2e)

In Jou, it returns `long`, and Jou's `long` is always 64 bit:

https://github.com/Akuli/jou/blob/2b33658b17239c68c84b034c33c2a7c145da43dd/stdlib/str.jou#L25-L27

See also #304.

PR #941, opened 2025-09-07
Add support for raspberry pi

Akuli:
This PR makes Jou work on 64-bit raspberry pis.

This does NOT solve #561, because this PR is about a raspberry pi, not a pi pico or other actually small/embedded device.


Akuli:
CI fail is unrelated.

PR #960, opened 2025-10-19
Bootstrap Jou with a Python script instead of old compiler written in C

### LINKED ISSUE #890: How to audit/trust this project?
In most open-source projects, if you want to make sure that the author is not a [Jia Tan](https://en.wikipedia.org/wiki/XZ_Utils_backdoor), you can simply read the code of the project.

But in Jou, reading all code in the repository is not enough because of how bootstrapping works. You also need to read many old versions of the Jou compiler:

https://github.com/Akuli/jou/blob/266051d7509c29a5d3defb15a1b43764c35de65e/bootstrap.sh#L19-L40

This also affects using Jou in other projects.

To alleviate this, we should probably make the above commit list shorter:
- To shorten the commit list once, we could write a new Jou compiler (in e.g. Python or C) to compile the main Jou compiler.
- To slow down the growth of the commit list, we could avoid using new features in the Jou compiler until N features have built up on a waiting list. This slows down the growth by a factor of N.

Akuli:
Fixes #890 

Practically, this means that:
- you need to have Python installed to get started with developing Jou (unless you are on Windows and you use `./windows_setup.sh --small`)
- setting up Jou is much faster than before (unless you download a Windows release from GitHub, that is unaffected).

TODO:
- [x] make sure the new bootstrap process works on Windows
- [x] make sure the new bootstrap process works on MacOS
- [x] make sure the new bootstrap process works on NetBSD
- [x] update macos instructions to say that you need to have python installed
- [x] update linux instructions to say that you need to have python installed
- [x] update netbsd instructions to say that you need to have python installed
- [x] update windows instructions to say that you need to have python installed if you don't use `--small`
- [x] update remaining documentation about the bootstrap process

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ Following the [instructions in the README](README.md#setup) is enough.
     Any reasonably new version of Python will work.
 3. Open Git Bash from the start menu.
     **You must use Git Bash** for running bash scripts such as `windows_setup.sh` and `runtests.sh`.
-4. Clone the project with Git Bash (or command prompt):
+4. Clone the project:
     ```
     cd Desktop
     git clone https://github.com/Akuli/jou
@@ -34,6 +34,8 @@ Following the [instructions in the README](README.md#setup) is enough.
     you can instead run `./windows_setup.sh --small`.
     Instead of downloading the full mingw64 (about 360MB download),
     this will get a minimal version of MinGW from a [release](#releases) of Jou (about 50MB download).
+    It will also take `jou.exe` from the release instead of [bootstrapping with a Python script](README.md#bootstrapping),
+    so you don't need to have Python installed if you use `--small`.
 6. Compile Jou:
     ```
     source activate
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -39,6 +39,9 @@ links.json
 # Perlin noise example. File may easily end up in examples folder or in project root.
 perlin_noise.ppm
 
+# Python stuff
```
Akuli:
```suggestion
# Python stuff (bootstrap transpiler is written in Python)
```

bootstrap.sh
```diff
@@ -3,50 +3,39 @@
 # The Jou compiler is written in Jou, but it doesn't help you very much if you
 # have nothing that can compile or run Jou code. That's why this script exists.
 #
-# This script takes old versions of Jou from Git history, starting with the
-# last commit that came with a compiler written in C. The previous version of
-# Jou is always used to compile the next.
+# This script uses `jou_to_c.py` to convert the Jou compiler's source code to C
```
Akuli: File was renamed

Akuli:
It finally works on windows :D

Took a while to debug. I'm so glad gdb is a thing even on windows... Turns out it was a stack overflow, because Windows has a smaller stack than Linux and the Python script creates horribly bad C code. I just cranked up the optimization level lol...

PR #962, opened 2025-10-20
When casting floats to integers, handle out-of-bounds values 

### LINKED ISSUE #872: float/double to integer cast UB?
https://github.com/Akuli/jou/blob/abd0f0b01b6da3a143b7ec895416f2faf913f4a9/tests/should_succeed/as.jou#L30-L33

I should figure out whether this is still UB. If it is, I should do something about this, because subtle ways to invoke UB are not fun.

Akuli:
Previously `1234.5 as byte` produced undefined behavior. Now it produces 255 because the value is clipped to min and max of byte (0 and 255).

Infinite float values behave like you would expect. NaN produces zero.

Fixes #872.

### DIFF
--- a/compiler/evaluate.jou
+++ b/compiler/evaluate.jou
@@ -38,9 +38,6 @@ def evaluate_boolean_and_or(expr: AstExpression*, and_or: byte*) -> int:
     return value.boolean as int
 
 
-# min_value and max_value are provided as methods on Type (see types.jou).
-
-
 # Returns False if the expression is not a simple compile-time constant.
 #
 # As the naming suggests, the type hint is only a hint and result type may differ.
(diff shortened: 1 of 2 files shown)


compiler/evaluate.jou
```diff
@@ -38,60 +38,7 @@ def evaluate_boolean_and_or(expr: AstExpression*, and_or: byte*) -> int:
...
-            match t.size_in_bits:
-                case 8:
-                    return -0x80
-                case 16:
-                    return -0x8000
-                case 32:
-                    # TODO: This is more difficult than it should be.
-                    return -(0x8000_0000 as int64)
-                case 64:
-                    # TODO: This is more difficult than it should be. Relies on overflow.
```
Akuli: Should preserve these comments

PR #963, opened 2025-10-20
Create stdlib/utf8.jou

Akuli:
This PR adds basic UTF-8 handling functions to Jou standard library. For example, you can now do `utf8_char_count(string)` to count the number of characters in a string. This is not same as the number of bytes, aka `strlen`, because of how UTF-8 works.

For example, `ö` is one character, but two bytes, so `strlen("ö")` returns 2:

```python
import "stdlib/io.jou"
import "stdlib/str.jou"
import "stdlib/utf8.jou"

def main() -> int:
    printf("%d\n", strlen("ö") as int)  # Output: 2
    printf("%d\n", utf8_char_count("ö") as int)  # Output: 1
    return 0
```

For now, the UTF-8 module only processes existing UTF-8 text. It can validate or decode UTF-8, but currently it has nothing for producing UTF-8. I might add that later if it is needed for something.

### DIFF
--- a/tests/should_succeed/utf8_test.jou
+++ b/tests/should_succeed/utf8_test.jou
@@ -151,7 +151,7 @@ def test_decoding_random_data() -> None:
 
     # It is very consistently valid about 10% of the time. I compared with Python.
     #printf("%f\n", valid_count / (num_iters as double))
-    if 0.08*num_iters <= valid_count and valid_count <= 0.12*num_iters:
+    if 0.07*num_iters <= valid_count and valid_count <= 0.12*num_iters:
         printf("random data check ok\n")  # Output: random data check ok
     else:
         printf("valid_count=%d num_iters=%d\n", valid_count, num_iters)


stdlib/utf8.jou
```diff
@@ -0,0 +1,97 @@
...
+# pass strings around and not worry about how UTF-8 works. But when you need
+# something more, this file likely does what you need. For the rare cases when
+# this file is not enough, please create an issue on GitHub.
+#
+# For example, this file can tell you that "ö" is one character, but it cannot
+# tell you that the name of the character is "LATIN SMALL LETTER O WITH DIAERESIS".
+
+
+# Check if b is a continuation byte. A UTF-8 multibyte (that is, non-ASCII)
+# character starts with a start byte and then has zero or more continuation
```
Akuli:
```suggestion
# character starts with a start byte and then has one or more continuation
```

stdlib/utf8.jou
```diff
@@ -0,0 +1,97 @@
...
+    if u <= 0x10ffff:
+        return 4
+    return -1  # too big, will never appear in text even if technically valid utf-8
+
+
+# Consumes one codepoint (character) from a UTF-8 string. For example, if
+# s = "örkki", then utf8_decode_one(&s) returns 246 and advances s so that it
+# appears as "rkki". Returns 0 without moving s on error.
+@public
+def utf8_decode_one(s: byte**) -> uint32:
```
Akuli:
```suggestion
def utf8_decode_char(s: byte**) -> uint32:
```

PR #965, opened 2025-10-22
fix error msg to say global or const

### LINKED ISSUE #964: Bad error message for missing const keyword
```python
keyStrUriSafe: byte* = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+-$"
```

This produces:

```
compiler error in file "a.jou", line 20: missing 'global' keyword when defining global variable
```

Then I try this:

```python
global keyStrUriSafe: byte* = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+-$"
```

And I get:

```
compiler error in file "a.jou", line 20: specifying a value for a global variable is not supported yet
```

The correct code is:

```python
const keyStrUriSafe: byte* = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+-$"
```

The first error message should say "global or const" instead of just "global". Then I would know that I can also try with `const`.

olkku45:
big pull request

Fixes #964

### DIFF
--- a/tests/syntax_error/missing_global_keyword.jou
+++ b/tests/syntax_error/missing_global_keyword.jou
@@ -1,1 +1,0 @@
-foo: int  # Error: missing 'global' or 'const' keyword when defining global variable


tests/syntax_error/missing_global_keyword.jou
```diff
@@ -1 +1 @@
-foo: int  # Error: missing 'global' keyword when defining global variable
+foo: int  # Error: missing 'global' or 'const' keyword when defining global variable
```
Akuli: Can you delete this test while you're here? It is the same exact test as `missing_global_keyword_type.jou`.

PR #976, opened 2025-10-26
Implement >> operator (bitwise shift right)

Akuli:
Example:

```python

import "stdlib/io.jou"

def main() -> int:
    n = 11
    printf("%d\n", n >> 1)  # Output: 5
    return 0
```

Shift left (<<) is more complicated, so for now I implemented only shifting right.

Checks a box in #879.

### DIFF
--- a/tests/should_succeed/bitwise_shift_right.jou
+++ b/tests/should_succeed/bitwise_shift_right.jou
@@ -124,7 +124,8 @@ def main() -> int:
     printf("%d\n", 97 >> 257)  # Output: 0
     printf("%d\n", 'a' >> 257)  # Output: 0
 
-    # Type hint on shift goes to the left side of shift
+    # Type hint on shift goes to the left side of shift, so compiler knows that
+    # the big number here is supposed to be a uint64, not int
     big_number: uint64 = 12345432123454321 >> 1
     printf("%llu\n", big_number)  # Output: 6172716061727160
 


compiler/typecheck/step3_function_and_method_bodies.jou
```diff
@@ -463,34 +474,39 @@ def check_binop(
...
-    do_implicit_cast(fom, lhs, cast_type, Location{}, NULL)
-    do_implicit_cast(fom, rhs, cast_type, Location{}, NULL)
-
-    if some_kind_of_compare:
-        return boolType
+    if op == AstExpressionKind.BitShiftRight:
+        # Special-cased, because right side type does not affect the type of result.
+        do_implicit_cast(fom, rhs, int64Type, Location{}, NULL)
+        return lhs.types.orig_type
     else:
```
Akuli: Let's write this with relying on the early return instead of `else` so that both the diff and resulting code are much more readable.

compiler/typecheck/step3_function_and_method_bodies.jou
```diff
@@ -1470,12 +1483,17 @@ def typecheck_statement(state: State*, stmt: AstStatement*) -> None:
             | AstStatementKind.InPlaceBitAnd
             | AstStatementKind.InPlaceBitOr
             | AstStatementKind.InPlaceBitXor
+            | AstStatementKind.InPlaceBitShiftRight
         ):
             targetexpr = &stmt.assignment.target
             valueexpr = &stmt.assignment.value
 
             targettype = typecheck_expression_not_void(state, targetexpr, NULL)
```
Akuli: Perhaps this should pass along the type hint, so that this works:

```python
big_number: uint64 = 12345432123454321 >> 1
```

compiler/typecheck/step3_function_and_method_bodies.jou
```diff
@@ -1503,6 +1521,9 @@ def typecheck_statement(state: State*, stmt: AstStatement*) -> None:
                 case AstStatementKind.InPlaceBitXor:
                     op = AstExpressionKind.BitXor
                     opname = "xor"
+                case AstStatementKind.InPlaceBitShiftRight:
+                    op = AstExpressionKind.BitShiftRight
+                    opname = "bruh2"
```
Akuli: Is this used anywhere?

compiler/parser.jou
```diff
@@ -118,6 +123,10 @@ def determine_the_kind_of_a_statement_that_starts_with_an_expression(
         return AstStatementKind.InPlaceBitOr
     if this_token_is_after_that_initial_expression.is_operator("^="):
         return AstStatementKind.InPlaceBitXor
+    if this_token_is_after_that_initial_expression.is_operator("<<="):
+        fail(this_token_is_after_that_initial_expression.location, "sorry, the bitwise '<<=' operator is not supported yet")
```
Akuli:
```suggestion
        fail(this_token_is_after_that_initial_expression.location, "sorry, the bitwise '<<=' operator is not supported yet (see issue #879 on GitHub)")
```

compiler/parser.jou
```diff
@@ -904,7 +913,7 @@ class Parser:
     # TODO: ban chaining different bitwise operations e.g. a ^ b & c
     def parse_expression_with_bitwise_ops(self) -> AstExpression:
         result = self.parse_expression_with_add()
-        while self.tokens.is_operator("&") or self.tokens.is_operator("|") or self.tokens.is_operator("^"):
+        while self.tokens.is_operator("&") or self.tokens.is_operator("|") or self.tokens.is_operator("^") or self.tokens.is_operator("<<") or self.tokens.is_operator(">>"):
```
Akuli:
```suggestion
        while (
            self.tokens.is_operator("&")
            or self.tokens.is_operator("|")
            or self.tokens.is_operator("^")
            or self.tokens.is_operator("<<")
            or self.tokens.is_operator(">>")
        ):
```

PR #994, opened 2025-11-01
Allow signed minimum values in const

### LINKED ISSUE #993: Smallest possible value of a signed type in const
Currently this doesn't work:

```
const INT_MIN: int = -0x8000_0000  # Error: value does not fit into int
```

The problem is that the compiler tries to evaluate `0x8000_0000`, which is indeed too large to fit into an `int`. But `-0x8000_0000` would fit, because of how signed integers work.

This is perhaps easier to understand with int8. Its valid values are from -128 to 127.

```
const INT8_MAX: int8 = 127  # works fine
const INT8_MIN: int8 = -128  # Error: value does not fit into int8
```

To fix this, the compiler should special-case these specific hard-coded values when it is evaluating the `-` operator at compile time.

There is no such problem with unsigned numbers.

Akuli:
Fixes #993

### DIFF
--- a/tests/should_succeed/const_signed_min.jou
+++ b/tests/should_succeed/const_signed_min.jou
@@ -2,12 +2,19 @@ import "stdlib/io.jou"
 
 # Making these work requires some special-casing in the compiler, that's why
 # they are tested here.
-const INT_MIN: int = -0x8000_0000
 const INT8_MIN: int8 = -128
-const INT64_MIN: int64 = -0x8000_0000_0000_0000
+const INT_MIN: int = -2147483648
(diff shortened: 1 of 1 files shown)


Akuli:
Lately I've done a few Jou bugfixes, including this one, like this:
1. Tell GitHub's free AI thing to fix the bug
2. Realize how horribly overcomplicated the AI's solution is and rewrite it myself entirely
3. Make a pull request

This seems to work well:
- Compared to doing everything myself, I quickly get an idea of whether the fix is as simple as I expect it to be, and I know where I need to make changes.
- Compared to doing everything with AI, I don't get bad code quality and questionable ownership of my project.

PR #1003, opened 2025-11-02
Compiler refactor: delete CompileState

Akuli:
Currently there are `CompileState` (old main.jou thing) and `CompilerState` (#1001). This PR deletes `CompileState`.

### DIFF
--- a/compiler/state.jou
+++ b/compiler/state.jou
@@ -8,7 +8,7 @@ import "./command_line_args.jou"
 
 @public
 class FileState:
-    path: byte*  # owned
+    path: byte*  # owned, unique, always set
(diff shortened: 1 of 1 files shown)


compiler/state.jou
```diff
@@ -1,9 +1,33 @@
...
+import "stdlib/list.jou"
+
+import "./ast.jou"
+import "./llvm.jou"
 import "./command_line_args.jou"
 
 
+@public
+class FileState:
+    path: byte*  # owned
```
Akuli:
```suggestion
    path: byte*  # owned, unique, always set
```

PR #1011, opened 2025-11-07
Add `typedef` statement

Akuli:
Example:

```python
import "stdlib/io.jou"

typedef Thingy = int

def main() -> Thingy:
    x: Thingy = 123
    printf("%d\n", x)  # Output: 123
    return 0
```

This is implemented so that `Thingy` is a new type, not just another name for `int`. This is a bit messy, but necessary. Consider this:

```python
x: int = 123
y: Thingy = 456

printf("%d\n", x as int)  # Warning: unnecessary cast from int to int
printf("%d\n", y as int)  # No warning
```

It is important that warnings work this way, because I want to use `typedef` for platform-specific things (32-bit support...). So it's important that you don't get a warning if you cast a `typedef` type to `int`, even if it happens to be already `int` on some platform. But it is important that you get a warning for unnecessary casts in general.

TODO:
- [ ] document typedefs
- [ ] mention typedefs in implicit casts documentation

### DIFF
--- a/compiler/typecheck/step2_populate_types.jou
+++ b/compiler/typecheck/step2_populate_types.jou
@@ -48,7 +48,7 @@ def handle_const(ft: FileTypes*, location: Location, c: AstConstStatement*) -> E
         fail(location, "cannot evaluate value of constant at compile time")
 
     # TODO: allow implicit casts?
-    if value.get_type().unwrap_typedef() != expected_type.unwrap_typedef():
+    if value.get_type().unwrap_typedef(&location) != expected_type.unwrap_typedef(&location):
         msg: byte[500]
         snprintf(
             msg, sizeof(msg),
@@ -117,7 +117,7 @@ def handle_signature(ft: FileTypes*, ast: AstFunctionOrMethod*, self_class: Type
         ft.add_function(&sig, usedptr)
 
     if sig.is_main_function():
-        if sig.return_type == NULL or sig.return_type.unwrap_typedef() != int_type(32):
+        if sig.return_type == NULL or sig.return_type.unwrap_typedef(&astsig.return_type.location) != int_type(32):
             fail(astsig.return_type.location, "the main() function must return int")
         if (
             sig.args.len != 0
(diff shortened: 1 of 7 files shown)


compiler/typecheck/step2_populate_types.jou
```diff
@@ -243,7 +241,17 @@ def typecheck_step2_populate_types(ast: AstFile*) -> List[ExportSymbol]:
                     exports.append(exp)
             case AstStatementKind.Class:
                 handle_class_members(&ast.types, &stmt.classdef)
-            case AstStatementKind.Enum | AstStatementKind.Link:
+            case AstStatementKind.TypeDef:
+                actual_type = type_from_ast(&ast.types, NULL, &stmt.typedef_statement.ast_type).unwrap_typedef()
```
Akuli: Should ensure that tests fail if I remove the `.unwrap_typedef()` at the end.

compiler/ast.jou
```diff
@@ -1254,3 +1261,21 @@ class AstEnumDef:
...
+    public: bool  # is it decorated with @public
+    used: bool  # used to detect unused typedefs
+
+    def print(self) -> None:
+        printf("typedef \"%s\" to be ", self.name)
+        self.ast_type.print(True)
+        putchar('\n')
+
+    def free(self) -> None:
+        self.ast_type.free()
```
Akuli: does it valgrind nicely? `./runtests.sh --valgrind typedef`

compiler/types.jou
```diff
@@ -101,8 +102,15 @@ class Type:
         func_ptr: FuncPtrType
         classdata: ClassData  # Class
         array: ArrayType  # Array
+        typedef_actual_type: Type*  # TypeDef
         enum_members: List[byte[100]]
 
+    def unwrap_typedef(self) -> Type*:
+        if self.kind == TypeKind.TypeDef:
+            assert self.typedef_actual_type != NULL
```
Akuli:
```suggestion
            assert self.typedef_actual_type != NULL
            assert self.typedef_actual_type.kind != TypeKind.TypeDef
```

Issue #1016, opened 2025-11-08
Compiler cache bug


Akuli:
Deleting `jou_compiled/` should never affect the results of compiling, but I ran into a case where it does.

To reproduce:
1. Check out commit deb0fe0
2. Delete all `jou_compiled` folders
3. Compile the Jou compiler with the bootstrap compiler (run `make`)
4. Apply the diff below using `git apply`
5. Compile the Jou compiler with the bootstrap compiler (run `make`)
6. Try `./jou examples/hello.jou`, it fails
7. Delete `compiler/jou_compiled` and `./jou`
8. Compile the Jou compiler with the bootstrap compiler (run `make`)
9. Try `./jou examples/hello.jou`, it succeeds

<details><summary>The diff</summary>

```diff
commit eb79848ed6d5487f6a4da445eb8d7f3b41a81395
Author: Akuli <akuviljanen17@gmail.com>
Date:   Sat Nov 8 04:37:58 2025 +0200

    why this broke? IT IS A COMPILER CACHE BUG!!!! must figure out later...

diff --git a/compiler/ast.jou b/compiler/ast.jou
index f8d3ea6..ac94326 100644
--- a/compiler/ast.jou
+++ b/compiler/ast.jou
@@ -11,6 +11,7 @@ import "stdlib/str.jou"
 import "stdlib/mem.jou"
 
 import "./errors_and_warnings.jou"
+import "./global_symbols.jou"
 import "./constants.jou"
 import "./types.jou"
 import "./types_in_ast.jou"
@@ -573,15 +574,12 @@ class AstAssertion:
 # const name: type = value
 @public
 class AstConstStatement:
-    name: byte[100]
+    symbol: GlobalSymbol
     type: AstType
     value: AstExpression
 
-    public: bool  # is it decorated with @public
-    used: bool  # used to detect unused constants
-
     def print(self) -> None:
-        printf("%s: ", self.name)
+        printf("%s: ", self.symbol.name)
         self.type.print(True)
         printf("\n")
 
@@ -591,6 +589,7 @@ class AstConstStatement:
         self.value.print_with_tree_printer(sub)
 
     def free(self) -> None:
+        self.symbol.free()
         self.type.free()
         self.value.free()
 
@@ -829,6 +828,56 @@ class AstStatement:
             case AstStatementKind.Pass | AstStatementKind.Break | AstStatementKind.Continue:
                 pass
 
+    def symbol(self) -> GlobalSymbol*:
+        match self.kind:
[...]

PR #1024, opened 2025-11-09
Compiler code cleanup: type-checking global variables

Akuli:
This makes global variables work more similarly to how `const` works since #1023. This way it will be easy to e.g. allow using `sizeof(some_global_var)` in `const` statements if I decide to do that later.

This also touches the hashing code, so #1016 may be affected in some way. I don't know yet.

### DIFF
--- a/compiler/typecheck/typecheck_global_symbols.jou
+++ b/compiler/typecheck/typecheck_global_symbols.jou
@@ -69,12 +69,11 @@ def typecheck_statement_with_symbol(jou_file: JouFile*, stmt: AstStatement*) -> 
 # This also finds things defined in the same file.
 #
 # Note: For special constants like "WINDOWS" this returns NULL, be aware!!!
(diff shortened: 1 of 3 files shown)


compiler/typecheck/typecheck_global_symbols.jou
```diff
@@ -61,16 +115,22 @@ def find_and_typecheck_constant(jou_file: JouFile*, name: byte*, out_ptr: Consta
...
-            typecheck_const(jou_file, stmt.location, &stmt.const_statement)
-        case GlobalSymbolStatus.TypeChecking:
-            fail(stmt.location, "value of constant depends on itself")
-        case GlobalSymbolStatus.Ready:
-            pass
+    typecheck_statement_with_symbol(jou_file, stmt)
 
     assert stmt.symbol().status == GlobalSymbolStatus.Ready
     if mark_used:
         stmt.symbol().used = True
```
Akuli:
```suggestion
```

PR #1030, opened 2025-11-09
Add error for classes that refer to each other + other compiler code cleanup

### LINKED ISSUE #299: segfault with cyclic-referencing classes
```
class Foo:
    bar: Bar

class Bar:
    foo: Foo
```

Expected result: a clear error message that explains what is wrong

If one of the two members is a pointer (e.g. `foo: Foo*`), this works as expected. You must use a pointer somewhere, just like in C, but if you don't, there should be a clear error message.

Akuli:
Recently I've been refactoring the type checking so that we type-check things when we need them, no earlier. This PR applies that to classes. Turns out that just doing it is enough to fix a very old compiler bug: Fixes #299.


Akuli:
I don't think the code is any "cleaner" than before :)

PR #1041, opened 2025-11-15
Add `enum_count` compiler built-in

### LINKED ISSUE #1038: Number of elements in an enum
I want to add a compiler builtin to get the number of elements in an enum. Maybe like this:

```python
enum Thing:
    Foo
    Bar
    Baz

def main() -> int:
    printf("%d\n", enum_count(Thing))  # Output: 3
    return 0
```

The reason is that in C, it is common to add a dummy "count" member to the end to figure out how many enum items you have. In fact, you can already do that in Jou:

```python
enum Thing:
    Foo
    Bar
    Baz
    Count

def main() -> int:
    printf("%d\n", Thing.Count as int)  # Output: 3
    return 0
```

but the problem is that if you now try to use a `Thing` in a `match` statement, the compiler will complain if you don't handle the `Count` member.

Akuli:
Example:

```python
import "stdlib/io.jou"

enum Operation:
    Add
    Subtract
    Multiply

def main() -> int:
    printf("%d\n", enum_count(Operation))  # Output: 3
    return 0
```

Fixes #1038

### DIFF
--- a/compiler/tokenizer.jou
+++ b/compiler/tokenizer.jou
@@ -20,10 +20,10 @@ def is_keyword(word: byte*) -> bool:
     match word with strcmp:
         case (
             "import" | "link"
-            | "def" | "declare" | "class" | "union" | "enum" | "enum_count" | "typedef" | "global" | "const"
+            | "def" | "declare" | "class" | "union" | "enum" | "typedef" | "global" | "const"
             | "if" | "elif" | "else" | "while" | "for" | "pass" | "break" | "continue" | "match" | "with" | "case" | "return"
             | "True" | "False" | "None" | "NULL" | "self"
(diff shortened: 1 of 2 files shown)


compiler/tokenizer.jou
```diff
@@ -20,7 +20,7 @@ def is_keyword(word: byte*) -> bool:
     match word with strcmp:
         case (
             "import" | "link"
-            | "def" | "declare" | "class" | "union" | "enum" | "typedef" | "global" | "const"
+            | "def" | "declare" | "class" | "union" | "enum" | "enum_count" | "typedef" | "global" | "const"
```
Akuli: Might make more sense after `array_count` than here

Issue #1045, opened 2025-11-16
More complicated array sizes


Akuli:
The Jou language has grown quite a lot since arrays were implemented. I think it's about time to update the array syntax to support new language features in array sizes.

Here are things that I want to do, but are currently not supported:
- [ ] `const` in array size
- [ ] `enum_count(Foo)` in array size
- [ ] `array_count(foo)` in array size
- [ ] `sizeof(foo)` in array size

Example:

```python
import "stdlib/mem.jou"

const MAX_PLAYERS: int = 123

enum Feature:
    Foo
    Bar

class Thing:
    x: int

def main() -> int:
    scores: int[MAX_PLAYERS]
    score_strings: byte*[array_count(scores)]

    features_enabled: bool[enum_count(Feature)]

    thing = Thing{x=1}
    thing_memory: byte[sizeof(thing)]
    memcpy(&thing_memory, &thing, sizeof(thing))

    return 0
```

The most difficult to implement will probably be `array_count(scores)` and `sizeof(thing)`, because now constant evaluation needs to know about local variables.

Issue #845, opened 2025-04-07
Bad error message for wrong array syntax


Akuli:
```python3
def foo(data: byte[]) -> int:
    pass
```

This should probably be `byte*`, perhaps together with a separate `length: int` parameter, or maybe `List[byte]`.

The compiler's error message for this is not very good:

```
compiler error in file "/home/akuli/jou/aa.jou", line 1: there is no generic class named 'byte'
```

Moosems:
This feels more misleading than not very good

PR #1055, opened 2025-11-19
Fix the compiler so it handles ambiguous `foo[bar]` type syntax

### LINKED ISSUE #1054: bool array with enum_count
It seems like `enum_count` still doesn't work in array sizes. Consider this code, taken from curses-klondike:

```python
class ParseState:
    program_name: byte*
    remaining_args: byte**  # never includes program name
    # TODO(Jou): This doesn't work yet. For now let's add much more than enough...
    #seen: bool[enum_count(Option)]
    seen: bool[10]
    result: Args
```

If I delete `seen: bool[10]` and uncomment the commented out code, it fails:

```
compiler error in file "src/args.jou", line 45: expected a type, got the 'enum_count' keyword
```

#1051 does not help with this.

Akuli:
The type `foo[bar]` can be either an array of `foo` whose length is `bar`, or a generic class `foo` with parameter `bar`. This pull request properly implements everything that the compiler needs to know about this ambiguity. If I did this right, it can now figure out the ambiguity in all possible cases.

Checks a couple boxes in #1045. Fixes #1054. Affects #845 but doesn't fix it.


compiler/tokenizer.jou
```diff
@@ -15,6 +15,7 @@ def is_identifier_or_number_byte(b: byte) -> bool:
 def is_operator_byte(c: byte) -> bool:
     return c != '\0' and strchr("=<>!.,()[]{};:+-*/&%|^~", c) != NULL
 
+@public
```
Akuli: nah
```suggestion
```

compiler/typecheck/common.jou
```diff
@@ -23,6 +25,75 @@ def plural_s(n: int64) -> byte*:
...
+            }
+        case AstTypeKind.Array:
+            # TODO: implement this
+            fail(type.location, "cannot evaluate array length at compile time")
+        case AstTypeKind.Generic:
+            # TODO: implement this
+            fail(type.location, "cannot evaluate array length at compile time")
+        case AstTypeKind.Pointer | AstTypeKind.FuncPtr | AstTypeKind.BuiltIn:
+            # "Foo*" or "funcptr(...) -> ..." or "int" is not a valid expression
+            # TODO: test this
```
Akuli: Tested now
```suggestion
```

compiler/typecheck/common.jou
```diff
@@ -23,6 +25,74 @@ def plural_s(n: int64) -> byte*:
         return "s"
 
 
+def array_type_from_ast(jou_file: JouFile*, item_type: Type*, length_ast: AstExpression*) -> Type*:
```
Akuli: Seems to be called in only one place, inline?

compiler/typecheck/common.jou
```diff
@@ -23,6 +25,74 @@ def plural_s(n: int64) -> byte*:
...
+# The syntax is ambiguous, and the parser assumes generic rather than array.
+# This function tries to interpret the type Foo[Bar] as an array.
+def handle_bad_generic(jou_file: JouFile*, generic: AstGenericType*, location: Location) -> Type*:
+    assert generic != NULL
+    name: byte* = generic.class_name
+    msg: byte[500]
+
+    item_type = find_and_typecheck_type(jou_file, name)
+    if item_type == NULL:
+        # TODO: make sure both error messages are tested
```
Akuli: Must do

tests/syntax_error/array_length_from_type.jou
```diff
@@ -0,0 +1,10 @@
+# This is a funny corner case. The compiler first interprets Foo[x*] as a
+# generic class, but then realizes that Foo is not generic and attempts to
+# interpret x* as an expression instead of a type.
+
+class Foo:
+    pass
+
+def blah() -> None:
+    x = 123
+    y: Foo[x*]  # Error: invalid syntax in array length
```
Akuli: Not the best error message, maybe:


```suggestion
    y: Foo[x*]  # Error: invalid array length
```

PR #1063, opened 2025-11-21
Make "jou --check" about 2 times faster

### LINKED ISSUE #1052: Compiler performance
I measured compiler performance in #1046 by compiling the Jou compiler itself. But I also have another big Jou project that is unforunately not open-source, at least not yet.

Here's a summary of my measurements, all on the same computer:

|Project|Compile from scratch|Compile fully cached (no LLVM stuff)|Linker only|
|----|---|---|----|
|Jou compiler|8.7 seconds|830ms|550ms|
|My other project|7.2 seconds|350ms|80ms|

From these, we can calculate how much time each thing takes:

|Project|LLVM|Linker|Everything else|
|----|---|---|----|
|Jou compiler|7.9 seconds|550ms|280ms|
|My other project|6.9 seconds|80ms|270ms|

So LLVM stuff is much slower than anything else, but in my other project, the linker runs faster so the "Everything else" time feels much more annoying when LLVM stuff is cached. I should optimize the compiler to do the "Everything else" stuff faster. I haven't thought much about compiler performance when making the compiler, so this is kind of expected :)

Akuli:
Fixes #1052 

This PR also affects a few other things:
- `starts_with` in stdlib/str.jou is now implemented as a simple for loop, not by calling `strncmp()`. The simple for loop is much faster when the second string is short and known at compile time, which is how `starts_with()` is almost always used.
- On systems other than Windows, `make kcachegrind` now profiles the code and opens the profiling results in kcachegrind. This is useful for finding slow functions.
- On systems other than Windows, `make hyperfine` now compares performance as shown in the below screenshot. This is useful for measuring whether the compiler became faster due to a code change, and how much.

screenshot-1763765431

### DIFF
--- a/compiler/tokenizer.jou
+++ b/compiler/tokenizer.jou
@@ -244,12 +244,14 @@ class Tokenizer:
 
     # TODO: Get rid of read_byte() / unread_byte()?
     def read_byte(self) -> byte:
-        c = *self.file_content_ptr
-        if c != '\0':
-            self.file_content_ptr++
-        if c == '\n':
-            self.location.lineno++
-        return c
+        match *self.file_content_ptr:
+            case '\0':
+                return '\0'
+            case '\n':
+                self.location.lineno++
+                return *self.file_content_ptr++
+            case _:
+                return *self.file_content_ptr++
 
     def unread_byte(self, b: byte) -> None:
         if b == '\0':


compiler/tokenizer.jou
```diff
@@ -152,102 +228,20 @@ def flip_paren(c: byte) -> byte:
...
-                self.last_3_bytes[0] = self.last_3_bytes[1]
-                self.last_3_bytes[1] = self.last_3_bytes[2]
-                self.last_3_bytes[2] = c
-                bad = detect_banned_whitespace_characters(self.last_3_bytes)
-                if bad != NULL:
-                    msg: byte[500]
-                    snprintf(msg, sizeof(msg), "source file contains Unicode whitespace character %s", bad)
-                    fail(self.location, msg)
-
+        c = *self.file_content_ptr
```
Akuli: Might as well use a `match` statement to be really clear

PR #1074, opened 2025-11-22
Add threading to compiler

Akuli:
This makes the compiler a lot faster. 

Compiling time basically consists of three things:
- Time spent in things that `jou --check` would do (recently [I made this 2 times faster](https://github.com/Akuli/jou/pull/1063))
- Time spent compiling object files with LLVM (this is by far the slowest)
- Time spent running the linker (we can't do much about this, except maybe change linker?)

This PR makes compiling object files with LLVM faster by doing it in multiple threads. For example, on my laptop with nothing cached, compiling the Jou compiler takes 15.4 seconds before this PR and 10.8 seconds with this PR. You would get more speedup on a computer with more CPU cores; my laptop only has two.

### DIFF
--- a/compiler/builders/llvm_builder.jou
+++ b/compiler/builders/llvm_builder.jou
@@ -813,16 +813,16 @@ def build_llvm_ir(state: ThreadLocalCompilerState*, jou_file: JouFile*) -> LLVMM
             case _:
                 pass
 
-#    for stmt = jou_file.ast.ptr; stmt < jou_file.ast.end(); stmt++:
-#        match stmt.kind:
-#            case AstStatementKind.FunctionDef:
-#                feed_ast_to_builder(&stmt.function, stmt.location, jou_file.is_main_file, &builder_wrapper)
+    for stmt = jou_file.ast.ptr; stmt < jou_file.ast.end(); stmt++:
+        match stmt.kind:
+            case AstStatementKind.FunctionDef:
+                feed_ast_to_builder(&stmt.function, stmt.location, jou_file.is_main_file, &builder_wrapper)
 #            case AstStatementKind.Class:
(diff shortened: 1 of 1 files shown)


compiler/command_line_args.jou
```diff
@@ -115,6 +118,15 @@ def check_args(args: CommandLineArgs*, optlevel_set: bool) -> None:
     if args.linker_flags != NULL and args.mode != CompilingMode.Run and args.mode != CompilingMode.CompileToFile:
         warn_about_meaningless_arg(args, "--linker-flags")
 
+    # Parallelism doesn't make sense if we won't invoke LLVM, because only LLVM
+    # stuff is parallelized.
+    #
+    # TODO: test this, and test the -j flag in general...
```
Akuli: Done


```suggestion
```

compiler/llvm.jou
```diff
@@ -263,24 +266,41 @@ enum LLVMRealPredicate:
...
+declare LLVMStructTypeInContext(C: LLVMContext*, ElementTypes: LLVMType**, ElementCount: int, Packed: int) -> LLVMType*
 @public
 declare LLVMArrayType(ElementType: LLVMType*, ElementCount: int) -> LLVMType*
 @public
-declare LLVMPointerType(ElementType: LLVMType*, AddressSpace: int) -> LLVMType*
+declare LLVMPointerTypeInContext(C: LLVMContext*, AddressSpace: uint32) -> LLVMType*
 @public
 declare LLVMDisposeMessage(Message: byte*) -> None
+#@public
+#declare LLVMModuleCreateWithName(ModuleID: byte*) -> LLVMModule*
```
Akuli: Should delete this and other similarly commented out things

Akuli:
Everything works for me locally, but according to CI, there are random failures that happen rarely on all platforms.

I am stuck.

PR #1081, opened 2025-11-24
Improve various error messages

Akuli:
Before:
- `print("hello")` --> `function 'print' not found`
- `import "stdlib/io"` --> `cannot import from "/home/akuli/jou/stdlib/io": No such file or directory`
- `global foo` in a function --> `global variables must be on top level, not e.g. inside a function`
- `var x = 1` --> `not a valid statement`
- `let x = 1` --> `not a valid statement`
- `else if` --> `expected ':' followed by a new line with more indentation, got the 'if' keyword`

After:
- `print("hello")` --> `function 'print' not found, try 'printf' from "stdlib/io.jou"`
- `import "stdlib/io"` --> `use "stdlib/io.jou" instead of "stdlib/io"`
- `global foo` in a function --> `you don't need to use 'global' inside a function or method to use a global variable`
- `var x = 1` --> `there is no 'var' keyword in Jou`
- `let x = 1` --> `there is no 'let' keyword in Jou`
- `else if` --> `use 'elif' instead of 'else if'`

Thank you @taahol for the laughs and helping me discover this :)

### DIFF
--- a/compiler/parser.jou
+++ b/compiler/parser.jou
@@ -1338,7 +1338,7 @@ class Parser:
             ):
                 fail(
                     self.tokens.location,
-                    "you don't need to use 'global' inside a function or method to use a global variable",
+                    "you don't need 'global' inside a function or method to use a global variable",
                 )
 
             self.check_top_level("global variables")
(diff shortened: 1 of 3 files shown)


tests/404/file_missing_extension.jou
```diff
@@ -0,0 +1 @@
+import "stdlib/io"  # Error: cannot import from "<joudir>/stdlib/io" (try "stdlib/io.jou" instead)
```
Akuli: I think this error message could be even better:


```suggestion
import "stdlib/io"  # Error: use "stdlib/io.jou" instead of "stdlib/io"
```

PR #1096, opened 2025-11-30
Makefile: check whether AArch64 targets exist

taahol:


Akuli:
@taahol What do you think? (GitHub doesn't let me ask you to review this.)

PR #1097, opened 2025-11-30
Link Jou compiler with -pthread

taahol:


compiler/main.jou
```diff
@@ -23,6 +23,7 @@ import "./typecheck/step2_class_members.jou"
 import "./typecheck/step3_function_and_method_bodies.jou"
 import "./types.jou"
 
+link "-pthread"
```
Akuli: Please move this to compiler/pthreads.jou.

Akuli:
Thanks for reporting and fixing :) I don't know what brings in pthreads on my systems and GitHub Actions.

PR #1104, opened 2025-12-06
Add `embed_file()` compiler built-in

### LINKED ISSUE #607: Including resource files in executable
In C, if you are programming a game and you have e.g. image files, it's difficult to bake them into the executable so that it doesn't open any files when your game starts. Jou should have a built-in way to do it.

Akuli:
For example, the following compiles to a program that includes the contents of Jou's README, and prints the README when it runs:

```python
import "stdlib/io.jou"

global readme = embed_file("./README.md")

def main() -> int:
    puts(readme)
    return 0
```

The type of the `readme` variable is `byte[11368]` where `11368` is the file size.

Fixes #607

### DIFF
--- a/compiler/typecheck/step3_function_and_method_bodies.jou
+++ b/compiler/typecheck/step3_function_and_method_bodies.jou
@@ -1116,7 +1116,6 @@ def typecheck_expression(state: State*, expr: AstExpression*, type_hint: Type*) 
                 and expr.operands[0].constant.kind == ConstantKind.PointerString
             ):
                 # sizeof("foo") should be 4
-                # TODO: this probably won't work with const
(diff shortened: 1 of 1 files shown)


compiler/constants.jou
```diff
@@ -7,9 +8,9 @@ import "stdlib/mem.jou"
 import "./types.jou"
 
 
-def print_string(s: byte*) -> None:
+def print_string(s: byte*, len: int64) -> None:
     putchar('"')
-    for i = 0; s[i] != '\0'; i++:
+    for i = 0; i < len; i++:
```
Akuli: These changes don't seem to be necessary.

PR #1110, opened 2025-12-10
Fix `i++` and `array[i]` when the type of `i` is a typedef

### LINKED ISSUE #1107: `typedef` doesn’t get recognized as a numeric type
```python
typedef t = int
def main() -> int:
    x: t = 42
    x--
    return 0
```
Fails with `compiler error in file "typedef-decrement.jou", line 4: cannot decrement a value of type t`

Akuli:
Fixes #1107 so that @taahol can continue his work on 32-bit support.

This PR is a somewhat messy "temporary" fix. I will eventually need to refactor the builder stuff so that no typedefs are fed into builders ever.


taahol:
new error: `the index inside [...] must be an integer, not intnative`

PR #1124, opened 2025-12-15
Add support for 32-bit systems

taahol:


stdlib/intnative.jou
```diff
@@ -0,0 +1,6 @@
+if IS32BIT:
```
Akuli: To do the 32-bit type-check we talked about, can you just change this to `if True:` and run `jou --check compiler/main.jou`?

taahol: Nyt en kyllä ihan ymmärtänyt kysymystä.

Akuli:
Huh, we ended up fixing the same things, but our fixes have small differences so this is full of little conflicts now :D

Akuli:
One approach would be to merge (or rebase) this with the latest `main` so that you always choose your version. Then we can compare our fixes as a nice GitHub diff and choose whatever looks best.

**Edit:** Let me actually do just that...

compiler/evaluate.jou
```diff
@@ -24,6 +24,7 @@ def get_special_constant(name: byte*) -> Constant*:
             SpecialConstant{name = "WINDOWS", constant = Constant{kind = ConstantKind.Bool, boolean = WINDOWS}},
             SpecialConstant{name = "MACOS", constant = Constant{kind = ConstantKind.Bool, boolean = MACOS}},
             SpecialConstant{name = "NETBSD", constant = Constant{kind = ConstantKind.Bool, boolean = NETBSD}},
+            SpecialConstant{name = "IS32BIT", constant = Constant{kind = ConstantKind.Bool, boolean = IS32BIT}},
```
Akuli: Nit: I would prefer `IS_32BIT`.

Akuli (commented):
Thanks again for working on this. This is quite close to mergeable :)

taahol:
Fails on Raspberry Pi:
```
doing stuff
Converting Jou code to C...
Compiling C code...
renamed 'stdlib/utf8_old.jou' -> 'stdlib/utf8.jou'
Compiling the same Jou code again with the compiler we just made...
compiler error in file "/home/taahol/jou/tmp/bootstrap_cache/024_0d4b4082f6569131903af02ba5508210b8b474d8/stdlib/limits.jou", line 18: value does not fit into int
make: *** [Makefile.posix:42: jou_bootstrap] Error 1

real    5m0,459s
user    4m39,755s
sys     0m8,476s
```

alpine.expect
```diff
@@ -0,0 +1,53 @@
...
+expect {
+	"/ # " {
+		send "cd \r"
+	}
+}
+expect {
+	"~ # " {
+		send "echo FAIL >status && echo 'iface eth0 inet dhcp' >/etc/network/interfaces && ifup eth0 && "
+		send "apk update && apk upgrade && TERM=dumb apk add bash clang llvm-dev make git grep patch && "
+		send "git clone https://github.com/Akuli/jou -b $env(GITHUB_HEAD_REF) && cd jou && "
```
Akuli: This can fail in a number of interesting ways. For example:
- If the branch name contains a space, this will not work.
- If an attacker decides the branch name, this has arbitrary code execution vulnerability. (Doesn't really matter currently because the permissions on PR workflows are very limited anyway.)

It would be better to copy the whole repository inside the VM (though maybe not with netcat...), see linux-aarch64 workflow.

alpine.expect
```diff
@@ -0,0 +1,53 @@
+set timeout 300
+
+spawn sudo kvm -m 4G -smp 2 -drive file=alpine-virt.iso,format=raw,media=cdrom -drive file=rootfs.qcow2 -nographic
```
Akuli: Do you know why this doesn't work for me locally? The VM boots fine but `ping google.com` doesn't work after `ifup eth0` step.

alpine.expect
Akuli: `$env(GITHUB_SHA)` is also not quite right; that will fail if someone else than one of us two makes a pull request, because branches from forks are not fetched automatically when you `git clone`.

taahol: That said, it would indeed be easy to copy the repo to `rootfs/` before the vm is created.

Akuli:
Maybe this is why it fails? Should be fixed in any case.

```
$ git grep 'pthread_t =' compiler/pthreads.jou 
compiler/pthreads.jou:typedef pthread_t = int64  # usually size of pointer, so this assumes 64-bit platform
```

vm/alpine.sh
```diff
@@ -0,0 +1,82 @@
...
+mv -v alpine-virt-*-x86.iso virt.iso
+
+mkdir rootfs
+tar xf rootfs.tar.gz -C rootfs
+
+git bundle create rootfs/root/jou.bundle --all
+
+truncate -s 4G rootfs.img
+/sbin/mkfs.ext4 rootfs.img -d rootfs
+qemu-img convert rootfs.img -O qcow2 rootfs.qcow2
```
Akuli: Is there a reason to convert to qcow2 instead of passing the raw image to qemu directly?

Makefile.posix
```diff
@@ -36,10 +36,20 @@ config.jou: Makefile.posix
...
+	echo 'const LLVM_HAS_AARCH64: bool = $(shell $(LLVM_CONFIG) --targets-built | grep -q AArch64 && echo True || echo False)' >> config.jou
+	echo '' >> config.jou
+	echo '@public' >> config.jou
+	echo 'const LLVM_HAS_ARM: bool = $(shell $(LLVM_CONFIG) --targets-built | grep -q ARM && echo True || echo False)' >> config.jou
+	echo '' >> config.jou
+	echo '@public' >> config.jou
+	echo 'const LLVM_HAS_X86: bool = $(shell $(LLVM_CONFIG) --targets-built | grep -q X86 && echo True || echo False)' >> config.jou
+	echo '' >> config.jou
+	echo '@public' >> config.jou
+	echo 'const JOU_TARGET: byte* = ""' >> config.jou
```
Akuli: If I understand this correctly, empty string means "use LLVM's default", but you can set this to work around the raspberry pi CPU problem? So you're going to add the "if grep raspberry pi" workaround here?

To future-proof this, a better name would be `JOU_NATIVE_TARGET`.

bootstrap.sh
```diff
@@ -238,6 +258,23 @@ function compile_next_jou_compiler() {
         echo "Deleting version check..."
         sed -i -e "/Found unsupported LLVM version/d" Makefile.*
 
+        if [ $number -le 23 ]; then
+            echo "Patching support for IS_32BIT special constant..."
+            sed -i -e '/NETBSD.as.int/a\
+        case "IS_32BIT":\
+            return IS_32BIT as int'$'\n' compiler/evaluate.jou
+            if [ $number -eq 17 ]; then
```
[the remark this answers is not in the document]
taahol: Please review in https://github.com/Akuli/jou/pull/1153.

Akuli:
As mentioned, I want to tweak a few things later but I'll merge now.

Thanks for working on this so patiently! And Merry Christmas :)

PR #1125, opened 2025-12-15
Add stdlib/intnative.jou and `intnative` typedef

Akuli:
This is a step towards 32-bit support (#304) taken from #1124. This PR defines the new `intnative` data type and applies it to existing code where appropriate. Currently all of our supported platforms are 64-bit, so `intnative` is always same as `int64` for now.

Once this PR has been merged, the latest main branch can be merged into @taahol's #1124 to make that pull request smaller.

### DIFF
--- a/compiler/parser.jou
+++ b/compiler/parser.jou
@@ -1,5 +1,6 @@
 import "stdlib/assert.jou"
 import "stdlib/ascii.jou"
+import "stdlib/intnative.jou"
 import "stdlib/list.jou"
 import "stdlib/str.jou"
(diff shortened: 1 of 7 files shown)


Akuli:
There is another unexpected problem with this: how to print the length of a list? You can't do `printf("%d\n", list.len as int)`, because on 32-bit systems the compiler would complain that the cast is unnecessary.

taahol:
Would `printf("%zd", ...)` be of some use?

Akuli:
Maybe. I don't know how cross-platform that is.

Issue #1126, opened 2025-12-16
Unnecessary cast warning is misleading


Akuli:
With 64-bit `intnative`:

```python
import "stdlib/io.jou"
import "stdlib/intnative.jou"

def main() -> int:
    x: intnative = 0
    printf("%lld\n", (x-1) as int64)
    return 0
```

Expected output: -1 with no errors
Actual output: -1 with `compiler warning for file "a.jou", line 6: unnecessary cast from int64 to int64`

If I do `x as int64`, there is no compiler warning because the compiler knows that the type of `x` is a typedef, hence `x` is likely to vary across platforms. But if I do `(x-1) as int64`, it loses all that information.

IMO the correct fix is to delete the unnecessary cast warning. It is not helpful.

Akuli:
Fixed in #1130

Issue #1057, opened 2025-11-19
Instantiation syntax doesn't work with typedef


Akuli:
```python
import "stdlib/list.jou"

typedef L = List[int]

def main() -> int:
    x = L{}
    return 0
```

Expected result: no error
Actual result: `compiler error in file "/home/akuli/jou/a.jou", line 6: the L{...} syntax is only for classes, but L is a class`

Thanks to @taahol for laughs and helping me discover this.

PR #1130, opened 2025-12-16
Delete unnecessary cast warning

Akuli:
```python
import "stdlib/io.jou"

def main() -> int:
    x = 0
    printf("%d\n", x as int)
    return 0
```

Output before: `compiler warning for file "a.jou", line 5: unnecessary cast from int to int` `0`

Output after this PR: `0`

There are two reasons for deleting the warning:
- Sometimes it complains about casts that **are** actually necessary. See the example in #1126. It would be very difficult to get rid of all these bugs.
- This warning makes `typedef` a LOT more difficult to implement, because the compiler attempts to track which types are typedefs and which are the actual underlying types. Without this warning, #1057 and many older similar bugs basically don't exist.

### DIFF
--- a/tests/should_succeed/as.jou
+++ b/tests/should_succeed/as.jou
@@ -38,7 +38,6 @@ def main() -> int:
     # the type is somehow platform-dependent. For example, if the type of `x`
     # is `intnative`, then `x as int` and `x as int64` should succeed without
     # warnings on 32-bit and 64-bit systems.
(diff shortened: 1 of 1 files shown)


Akuli:
I found yet another case where the warning is wrong: `sizeof("hello" as byte*)` produces 8 and complains about unnecessary cast, but `sizeof("hello")` produces 6.

PR #1132, opened 2025-12-16
Use, document and test `%zd` in printf format strings

### LINKED ISSUE #1128: How to print an `intnative`?
> Would `printf("%zd", ...)` be of some use?

_Originally posted by @taahol in https://github.com/Akuli/jou/issues/1125#issuecomment-3658217015_
            

#1125 adds the `intnative` data type along with documentation that recommends `printf("%d\n", foo as int)` or `printf("%lld\n", foo as int64)` to print it. While these work, a simple `%zd`/`%zu` would be more convenient to use.

We should figure out whether `%zd` and `%zu` work on all platforms that Jou supports (that is, does CI succeed or fail if you use them). If they work, we should update the documentation to recommend them, and also add a comment about printing to stdlib/intnative.jou.

Akuli:
Fixes #1128

### DIFF
--- a/doc/types.md
+++ b/doc/types.md
@@ -51,8 +51,11 @@ use `u` for unsigned, and add `ll` for 64-bit values or `z` for native-size valu
 
 The `ll` in `%lld` and `%llu` is short for "long long",
 which is basically C's way to say "64-bit number".
-Do not use `%ld` or `%lu`, because they are like `%lld` and `%llu`,
-except that on 64-bit Windows they print 32-bit values.
+
+There is also `%ld` and `%lu`,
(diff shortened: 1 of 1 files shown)


compiler/uvg.jou
```diff
@@ -159,7 +159,7 @@ class Uvg:
 
         actual_varname: byte[100]
         if varname == NULL:
-            sprintf(actual_varname, "$%lld", self.varnames.len as int64)
+            sprintf(actual_varname, "$%zd", self.varnames.len as int64)
```
Akuli:
```suggestion
            sprintf(actual_varname, "$%zd", self.varnames.len)
```

PR #1155, opened 2025-12-23
Create vm/wait_for_string.sh script

Akuli:
In `netbsd.sh`, there are a couple places where I need to run netcat until a certain output appears. This has turned out to be more difficult than I expected it to be, so I added a script to automate that.

The script looks quite straight-forward, but it was surprisingly difficult to get right, because:
- `sed`, `grep`, `awk` etc process input line by line, so they don't do anything until a `\n` appears in the input and cannot be used to look for a shell's prompt
- we can't run the netcat command in the background (as in `&`), because backgrounded processes don't receive input the same way as the "main" process.

### DIFF
--- a/vm/README.md
+++ b/vm/README.md
@@ -30,7 +30,7 @@ Here are some guidelines for working on the scripts:
 - Use `set -e -o pipefail`. If this gets in your way, use `|| true` only in the
   places that need it. This way failures are not silenced except where you
   expect to get failures.
-- Use the [`wait_for_string.sh`](./wait_for_string.sh) script instead the
+- Use the [`wait_for_string.sh`](./wait_for_string.sh) script instead of the
   `expect` program, for two reasons:
     - Jou developers already need to know at least 4 programming languages, and
       I don't really want to add Tcl to the mix.
--- a/vm/netbsd.sh
+++ b/vm/netbsd.sh
@@ -102,8 +102,7 @@ if ! [ -f key ] || ! timeout 5 $ssh echo hello; then
     if ! [ -f key ] || ! timeout 5 $ssh echo hello; then
         echo "ssh doesn't work. Let's set it up using serial port."
         (yes || true) | ssh-keygen -t ed25519 -f key -N ''
-        # This was a bit tricky to set up. It kills netcat when the serial output
-        # mentions "ALLDONENOW" and also displays all output for debugging.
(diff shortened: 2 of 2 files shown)


vm/netbsd.sh
```diff
@@ -96,20 +96,15 @@ if ! [ -f key ] || ! timeout 5 $ssh echo hello; then
...
-        sleep 5
-    done
+    echo "Waiting for VM to boot..."
+    echo | ../wait_for_string.sh 'login:' nc localhost 4444
     echo "Checking again if ssh works..."
     if ! [ -f key ] || ! timeout 5 $ssh echo hello; then
         echo "ssh doesn't work. Let's set it up using serial port."
         (yes || true) | ssh-keygen -t ed25519 -f key -N ''
         # This was a bit tricky to set up. It kills netcat when the serial output
         # mentions "ALLDONENOW" and also displays all output for debugging.
```
Akuli:
```suggestion
        # Log in as root and set up ssh key
```

vm/README.md
```diff
@@ -30,8 +30,11 @@ Here are some guidelines for working on the scripts:
 - Use `set -e -o pipefail`. If this gets in your way, use `|| true` only in the
   places that need it. This way failures are not silenced except where you
   expect to get failures.
-- Do not use the `expect` program. It is meant for this kind of thing, but it
-  is an unnecessary dependency.
+- Use the [`wait_for_string.sh`](./wait_for_string.sh) script instead the
```
Akuli:
```suggestion
- Use the [`wait_for_string.sh`](./wait_for_string.sh) script instead of the
```

vm/wait_for_string.sh
```diff
@@ -0,0 +1,82 @@
...
+# The following example demonstrates how slow this script is (it is pretty
+# slow, but that doesn't matter for our use cases):
+#
+#   $ vm/wait_for_string.sh hi base64 /dev/urandom
+#
+# Implementing this is more difficult than you would expect, because many tools
+# (sed, grep, awk, ...) don't process any input until a newline character
+# appears, and forcing them to process byte by byte instead of line by line is
+# difficult. Performance is horrible because we work one byte a time, but it's
+# fine for this use case.
```
Akuli:
```suggestion
# not really a thing.
```

PR #1169, opened 2025-12-29
32-bit and `intnative` documentation improvements

### LINKED ISSUE #1150: intnative in Jou tutorial
Currently doc/tutorial.md uses a lot of `%lld` in places where it should use `%zd`. It should not only use more `%zd`, but explain clearly why and when it is used with links to more documentation.

### LINKED ISSUE #1166: doc/types.md says that size of any pointer is 8 bytes
This is only true on 64-bit platforms.

Akuli:
- Document 32-bit/64-bit and `intnative` in Jou tutorial.
- Do not say that a pointer is always 8 bytes.

Fixes #1150 
Fixes #1166

### DIFF
--- a/doc/tutorial.md
+++ b/doc/tutorial.md
@@ -452,7 +452,7 @@ but with various annoyances fixed, and of course, with Python's simple syntax.
 ## 32-bit vs 64-bit
 
 As you probably know, most computers are 64-bit.
-It basically means that the size of a [pointer](#pointers) is 64 bits,
+It basically means that the size of any [pointer](#pointers) is 64 bits,
 and in general, the computer is good at doing math with 64-bit numbers.
 
 The following program prints the size of a pointer.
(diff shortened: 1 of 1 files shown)


doc/tutorial.md
```diff
@@ -449,6 +449,73 @@ The purpose of Jou is to be a lot like C,
 but with various annoyances fixed, and of course, with Python's simple syntax.
 
 
+## 32-bit vs 64-bit
+
+As you probably know, most computers are 64-bit.
+It basically means that the size of a [pointer](#pointers) is 64 bits,
```
Akuli:
```suggestion
It basically means that the size of any [pointer](#pointers) is 64 bits,
```

doc/tutorial.md
```diff
@@ -449,6 +449,73 @@ The purpose of Jou is to be a lot like C,
...
+You can use the `int64` and `uint64` types even on 32-bit systems.
+This is always possible, because the compiler can use two 32-bit numbers to fake a 64-bit number:
+
+![A funny image that illustrates how 64-bit numbers work on 32-bit computers](images/64bit-meme-small.jpg)
+
+
+## `intnative`
+
+Many things in [C's standard library](#cs-standard-library-libc)
+are `int` on 32-bit systems and `int64` in 64-bit systems.
```
Akuli:
```suggestion
are `int` on 32-bit systems and `int64` on 64-bit systems.
```

PR #1172, opened 2025-12-29
runtests: replace nproc with getconf, don't skip one advent of code file

taahol:


runtests.sh
```diff
@@ -235,11 +235,6 @@ function should_skip()
         fi
     fi
 
-    # This file is very slow compared to other advent of codes
-    if [ $joufile = examples/aoc2023/day24/part2.jou ]; then
-        return 0
-    fi
```
Akuli: This can stay as far as I can tell.

runtests.sh
```diff
@@ -301,8 +296,8 @@ function run_test()
     fi
 }
 
-# How many tests to run in parallel? Fall back to 2 if nproc fails or doesn't exist.
-num_tests_in_parallel=$(nproc 2>/dev/null || echo 2)
+# How many tests to run in parallel? Fall back to 2 if getconf fails or doesn't exist.
+num_tests_in_parallel=$(getconf _NPROCESSORS_ONLN 2>/dev/null || echo 2)
```
Akuli: We don't really need a fallback, this should work on all systems we support:
```suggestion
# How many tests to run in parallel?
num_tests_in_parallel=$(getconf _NPROCESSORS_ONLN)
```

runtests.sh
Akuli: Nevermind, the test is pretty fast...

PR #1174, opened 2025-12-30
Set up Alpine Linux VM to run tests on armv6

Akuli:
The purpose is to reproduce the bug that #1154 fixed.

This was difficult for many reasons:
- Running armv6 on qemu is difficult. You can't just download a disk image and boot it, like you can on other platforms, because most disk images assume an early raspberry pi but qemu doesn't support all of its hardware. You instead need to get the kernel separately from https://github.com/dhruvvyas90/qemu-rpi-kernel.
- alpine's minirootfs is very... uh, minimal. It doesn't have an init.
- I configured GitHub Actions to do all linux VMs from the same file, so that there would be slightly less copy/pasta.


taahol:
What if we cross-compile and run the test suite with `qemu-user`? It might be easier to set up.

PR #1180, opened 2025-12-30
Run tests on armv6 with Raspberry Pi OS

Akuli:
This adds a VM that reproduces a bug that @taahol worked around: `llvm-config` reports `arm`, but the machine is `armv6`.

```
pi@raspberrypi:~ $ uname -mp
armv6l unknown
pi@raspberrypi:~ $ llvm-config-19 --host-target
arm-unknown-linux-gnueabihf
pi@raspberrypi:~ $ 
```

PR #1181, opened 2025-12-31
VM scripts refactor: separate keygen and ssh into new scripts

Akuli:
This PR creates two new scripts that do things that all VM scripts need to do:
- `vm/keygen.sh` runs `ssh-keygen` if ssh key doesn't exist
- `vm/ssh.sh` connects to any currently running VM with ssh.

The main reason for this refactor is to replace the following...

1. Open `vm/foo.sh`
2. Ctrl+F `ssh=`
3. Copy the command in quotes
4. `cd vm/foo-bar/`
5. Paste command to terminal and run it

...with the following:

1. Run `vm/ssh.sh`

In other words, it is now very easy to connect to a VM: no matter what VM you're running, `vm/ssh.sh` should just work.

### DIFF
--- a/vm/README.md
+++ b/vm/README.md
@@ -22,8 +22,8 @@ Here are some guidelines for working on the scripts:
   system, create a VM, set up the operating system in the VM, install packages
   and copy Jou into the VM. Once the script has completed, ssh should be
   available on port 2222 of localhost.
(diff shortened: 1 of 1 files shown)


Akuli:
This will conflict with #1180.

PR #1183, opened 2025-12-31
Use traditional netcat in vm scripts

### LINKED ISSUE #1182: How to run vm scripts locally?
This doesn’t really work:
```
$ vm/alpine.sh x86 echo hello
mkdir: skapade katalog 'vm/alpine-x86'
Downloading alpine linux...
mkdir: skapade katalog '../downloads'
../download.sh: downloading https://dl-cdn.alpinelinux.org/v3.23/releases/x86/alpine-virt-3.23.2-x86.iso
[...]
2025-12-31 10:18:48 (11,2 MB/s) - ”alpine-virt-3.23.2-x86.iso” sparades [53477376/53477376]

../download.sh: verifying...
'../downloads/alpine-virt-3.23.2-x86.iso' -> './alpine-virt-3.23.2-x86.iso'
../download.sh: done!
Creating ssh key...
Generating public/private ed25519 key pair.
[...]
Creating disk...
Booting temporary VM to install alpine...
Waiting for temporary VM to boot so we can install alpine...
Ncat: Connection refused.
Logging in to temporary VM...
Installing alpine...
u
```

Akuli:
Fixes #1182 

@taahol Can you try this and see if it works for you?

### DIFF
--- a/.github/workflows/linux-aarch64.yml
+++ b/.github/workflows/linux-aarch64.yml
@@ -42,7 +42,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0  # Fetch the whole Git history for bootstrapping
-      - run: sudo apt update && sudo apt -y install --no-install-recommends qemu-system-aarch64 qemu-efi-aarch64 ipxe-qemu
+      - run: sudo apt update && sudo apt -y install --no-install-recommends qemu-system-aarch64 qemu-efi-aarch64 ipxe-qemu netcat-traditional
       - uses: actions/cache@v5
         with:
           path: vm/alpine-aarch64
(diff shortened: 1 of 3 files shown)


Akuli:
Huh, it gets stuck even when login is available... I'll look into why.

Akuli:
I see now. The problem is that `ncat` closes the connection when its stdin is closed.

I'll change the scripts to not use `ncat`.

Akuli:
Now it seems to work for me. How about you?

Akuli:
I'm not not sure if other distros have `nc.traditional`, but they probably have `ncat`. So I might actually prefer `ncat --no-shutdown`. But that shall be another PR.

PR #1187, opened 2026-01-02
Compiler refactor: simplify main() handling

Akuli:
The `main()` function is special-cased in a few places. This PR doesn't change the behavior of the compiler at all, but makes the code easier to understand.

### DIFF
--- a/tests/404/main_function.jou
+++ b/tests/404/main_function.jou
@@ -1,0 +1,3 @@
+# Output: compiler error in file "tests/404/main_function.jou": missing `main` function to execute the program
+def test() -> None:  # Warning: function 'test' defined but not used
+    _ = "lol no main func"
--- a/tests/404/main_function_imported.jou
+++ b/tests/404/main_function_imported.jou
@@ -1,0 +1,13 @@
+# If an imported file has a main() function, it is ignored.
+#
+# This is useful in large Jou projects. For example, large_project/json_utils.jou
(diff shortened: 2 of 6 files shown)


compiler/typecheck/common.jou
```diff
@@ -256,7 +256,7 @@ def typecheck_signature(jou_file: JouFile*, ast: AstFunctionOrMethod*, self_clas
     else:
         sig.return_type = type_from_ast(jou_file, self_class, &astsig.return_type)
 
-    if sig.is_main_function():
+    if self_class == NULL and strcmp(sig.name, "main") == 0:
```
Akuli: Should probably add a test with method named `main`. It shouldn't cause problems.

compiler/main.jou
```diff
@@ -26,7 +26,10 @@ import "./types.jou"
 
 def defines_main(jou_file: JouFile*) -> bool:
     for stmt = jou_file.ast.ptr; stmt < jou_file.ast.end(); stmt++:
-        if stmt.kind == AstStatementKind.FunctionDef and stmt.symbol().signature.is_main_function():
+        if (
+            stmt.kind == AstStatementKind.FunctionDef
+            and strcmp(stmt.symbol().signature.name, "main") == 0
```
Akuli: This could be tested more too. These shouldn't be enough:
- main in imported file
- main method in some class
- global variable named main

PR #1192, opened 2026-01-03
Make methods private by default

### LINKED ISSUE #1077: Private methods
There are 3 reasonable ways how methods could be private/public:
1. Methods are always public. If you get an instance of Foo from somewhere, you can always call `foo.bar()` and the compiler will make it work.
2. Methods can be private to the class, meaning that you can only call them inside the class. Very traditional, can cause annoying "how do I access this from here" problems.
3. Methods can be private to file, so that other files cannot call them, just like functions. Rust does this.

Currently Jou uses option 1. 

I think maybe Jou should switch to option 3. It would be easy to implement in the compiler, and closer to the "methods are just functions with syntactic sugar" idea.

Akuli:
Fixes #1077 

This is backwards incompatible: you will need to put `@public` to all methods that are used in other files, and you will need to import the file that defines a method to be able to call it.

For example, if `foo.jou` has `def do_thing(self) -> None:`, and `bar.jou` wants to use the `do_thing()` method, you will need `@public` above the `def do_thing` line in `foo.jou` **and** `import "./foo.jou"` or similar in `bar.jou`.

This PR contains a horrible error message for "method not found". As the error message admits, it's bad and it will be removed. The following code:

```python
class Foo:
    pass

def thing() -> None:
    f = Foo{}
    f.bar()
```

...produces this error: `compiler error in file "a.jou", line 6: method 'bar' not found in class Foo. The problem is ONE of the following: 1) you didn't import the file that defines class Foo, 2) the 'bar' method is private, 3) there really is no such method in class Foo. I know that this is a bad compiler error message, and I will later split this into 3 different error messages.`

I didn't implement a proper error message right away, because that requires the compiler to know what private symbols would be avaliable if they were public, and that's really a new feature.

### DIFF
--- a/tests/should_succeed/import_generic_class.jou
+++ b/tests/should_succeed/import_generic_class.jou
@@ -1,11 +1,16 @@
+# There was a bug that caused generic classes to not appear as generic when imported.
+
 import "stdlib/io.jou"
+import "./imported/generic.jou"
 
-import "./imported/point_factory.jou"
-import "./imported/bar.jou"  # Contains definition of Point. Needed to access Point's methods.
-
+class Foo:
+    pair: Point2D[int]
 
 def main() -> int:
-    p = make_point()
-    sum = p.get_sum()
-    printf("%d\n", sum)  # Output: 3
+    point = Point2D[int]{x = 1, y = 2}
+    printf("%d %d\n", point.x, point.y)  # Output: 1 2
+
+    f = Foo{pair = point}
+    printf("%d %d\n", f.pair.x, f.pair.y)  # Output: 1 2
+
     return 0
--- a/tests/should_succeed/import_method.jou
+++ b/tests/should_succeed/import_method.jou
@@ -1,16 +1,11 @@
-# There was a bug that caused generic classes to not appear as generic when imported.
-
 import "stdlib/io.jou"
-import "./imported/generic.jou"
 
-class Foo:
-    pair: Point2D[int]
+import "./imported/point_factory.jou"
+import "./imported/bar.jou"  # Contains definition of Point. Needed to access Point's methods.
+
 
 def main() -> int:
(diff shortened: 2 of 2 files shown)


compiler/ast.jou
```diff
@@ -1329,6 +1342,7 @@ class AstFunctionOrMethod:
     body: List[AstStatement]
     locals: List[LocalVariable]  # populated in typecheck, empty before typecheck
     inline: bool  # is it decorated with @inline
+    containing_class_def: AstStatement*  # only for methods, the class definition where it is
```
Akuli: Is this needed?

compiler/parser.jou
```diff
@@ -1504,10 +1501,17 @@ class Parser:
...
 
-        symbol = GlobalSymbol{name = signature.name, kind = GlobalSymbolKind.Function}
-        if strcmp(symbol.name, "main") == 0 and self.is_main_file:
-            # Implicitly public main function
-            symbol.public = True
+        if is_method:
+            symbol = GlobalSymbol{kind = GlobalSymbolKind.Method}
+            assert self.containing_class_name[0] != '\0'
+            # Name the symbol e.g. "List.append"
+            # TODO: name might not fit...
```
Akuli: Should fix

compiler/parser.jou
```diff
@@ -288,6 +284,7 @@ class Parser:
     is_main_file: bool
     loop_vars: List[byte*]  # NULL means a loop that doesn't look like "for foo = bar; ..."
     what_are_we_parsing: WhatAreWeParsing
+    containing_class_name: byte[100]
```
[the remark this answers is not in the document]
Akuli: Maybe not, but this is a simple way to set the names of method symbols. Let's keep it.

PR #1195, opened 2026-01-03
Move ARMv6 target workaround to Makefile.posix

taahol:
Fixes: https://github.com/Akuli/jou/issues/1193

### DIFF
--- a/Makefile.posix
+++ b/Makefile.posix
@@ -43,13 +43,14 @@ config.jou: Makefile.posix
 	echo '' >> config.jou
 	echo '@public' >> config.jou
 	echo 'const LLVM_HAS_X86: bool = $(shell $(LLVM_CONFIG) --targets-built | grep -q X86 && echo True || echo False)' >> config.jou
+# LLVM seems to be misconfigured on Raspberry Pi OS so that it compiles to ARMv7 by default even if the CPU is ARMv6 (e.g. RasPi 1).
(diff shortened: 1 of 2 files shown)


Akuli:
On Windows, this...

```
const LLVM_HAS_AARCH64: bool = False
```

...was turned into this with `sed`:

```
const True: bool = False
```


I deleted that.

Makefile.posix
```diff
@@ -45,11 +45,11 @@ config.jou: Makefile.posix
 	echo 'const LLVM_HAS_X86: bool = $(shell $(LLVM_CONFIG) --targets-built | grep -q X86 && echo True || echo False)' >> config.jou
 	echo '' >> config.jou
 	echo '@public' >> config.jou
-	echo 'const JOU_TARGET: byte* = ""' >> config.jou
+	echo 'const JOU_TARGET: byte* = "$(shell $(LLVM_CONFIG) --host-target | ((uname -mp | grep -q armv6) && sed -e "s/^arm-/armv6-/" || cat))"' >> config.jou
```
Akuli: I don't like how this sets `JOU_TARGET` even if there's no need to set it. Can we leave it empty when the default is fine?

Something like this pseudo code maybe: `(check for armv6) && (host target | grep ^arm- | sed)`

taahol: De gustibus non est disputandum.

PR #1206, opened 2026-01-05
Replace the horrible method not found error message with 3 separate errors

Akuli:
Instead of one really long error message that lists 3 different possibilities that could cause it, let's just have 3 different error messages.

This was a bit of a hack to implement. It may be possible to clean this up in the future if I add some general way to look up private symbols of imported files. (That would be used only for error messages.)

### DIFF
--- a/compiler/typecheck/step3_function_and_method_bodies.jou
+++ b/compiler/typecheck/step3_function_and_method_bodies.jou
@@ -651,7 +651,7 @@ def fail_with_method_not_found(
         ):
             # Method exists
             if stmt.symbol().public:
-                # TODO: generate something that can be copy/pasted to become an import?
+                # TODO: generate something that can be copy/pasted to write an import statement?
                 snprintf(
                     msg, sizeof(msg),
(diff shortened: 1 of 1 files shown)


compiler/typecheck/step3_function_and_method_bodies.jou
```diff
@@ -629,6 +629,49 @@ def nth(n: int) -> byte[100]:
...
+    # Is the method in the class definition?
+    for stmt = class_ast.classdef.body.ptr; stmt < class_ast.classdef.body.end(); stmt++:
+        if (
+            stmt.kind == AstStatementKind.MethodDef
+            and strcmp(stmt.method.ast_signature.name, method_name) == 0
+            and strcmp(stmt.location.path, state.jou_file.path) != 0
+        ):
+            # Method exists
+            if stmt.symbol().public:
+                # TODO: generate something that can be copy/pasted to become an import?
```
Akuli:
```suggestion
                # TODO: generate something that can be copy/pasted to write an import statement?
```

PR #1215, opened 2026-01-19
Fix unwanted implicit integer conversion behavior

### LINKED ISSUE #1214: Implicit integer conversions behave weirdly
```python
import "stdlib/io.jou"
import "stdlib/limits.jou"

def main() -> int:
    minutes = 3
    if minutes <= UINT32_MAX / 60:
        printf("ok\n")
    else:
        printf("wat?\n")
    return 0
```

Expected result: `ok`
Actual result: `wat?`

I think `UINT32_MAX` is implicitly converted to `int`.

olkku45:
Fixes #1214.
Note: when running 'make' in the repository with this change, the compiler issues a warning:
```
compiler warning for file "compiler/typecheck/step3_function_and_method_bodies.jou", line 352: 
    function 'cast_type_of_integer_binop' doesn't seem to return a value in all cases
```


olkku45:
This seems to cause many issues.

PR #1223, opened 2026-02-07
Support arrays in compile-time values

### LINKED ISSUE #1222: Evaluating arrays at compile time
I often end up doing something like this: ([taken from curses-klondike](https://github.com/Akuli/curses-klondike/blob/7a434ae2d1c7af8a0176a0a3b2c023a7021936c2/src/help.jou#L39-L50))

```python
# TODO(Jou): use a `const` when that becomes possible
def get_picture_lines() -> byte*[8]:
    return [
        "╭──╮╭──╮    ╭──╮╭──╮╭──╮╭──╮",
        "│  ││  │    │ foundations  │",
        "╰──╯╰──╯    ╰──╯╰──╯╰──╯╰──╯",
        "  │   ╰─── discard          ",
        "  ╰─── stock                ",
        "╭──╮╭──╮╭──╮╭──╮╭──╮╭──╮╭──╮",
        "│         tableau          │",
        "╰──╯╰──╯╰──╯╰──╯╰──╯╰──╯╰──╯",
    ]
```

Akuli:
Fixes #1222

This PR is initially implemented with AI, but as usual, I'll probably rewrite all of it by hand as I review it.

### DIFF
--- a/compiler/llvm.jou
+++ b/compiler/llvm.jou
@@ -358,6 +358,7 @@ declare LLVMConstInt(IntTy: LLVMType*, N: int64, SignExtend: int) -> LLVMValue*
 declare LLVMConstRealOfString(RealTy: LLVMType*, Text: byte*) -> LLVMValue*
 @public
 declare LLVMConstStringInContext(C: LLVMContext*, Str: byte*, Length: uint32, DontNullTerminate: int) -> LLVMValue*
+# TODO: Use LLVMConstArray2 (new in LLVM 17) when we no longer support LLVM 16 and older
 @public
 declare LLVMConstArray(ElementTy: LLVMType*, ConstantVals: LLVMValue**, Length: uint32) -> LLVMValue*
 @public


compiler/constants.jou
```diff
@@ -91,6 +93,11 @@ class Constant:
                 printf("pointer string ")
                 print_string(self.pointer_string)
                 printf("\n")
+            case ConstantKind.Array:
+                # Ideally we would print the elements, but each constant would
+                # print a newline at the end, and I'm not sure if something
+                # relies on this thing outputting just one line.
+                printf("array of %d elements\n")
```
Akuli:
```suggestion
                printf("array of %zd elements\n", self.array_elements.len)
```

compiler/constants.jou
```diff
@@ -99,6 +106,10 @@ class Constant:
                 free(self.array_of_bytes.ptr)
             case ConstantKind.PointerString:
                 free(self.pointer_string)
+            case ConstantKind.Array:
+                for i = 0; i < self.array_elements.len; i++:
+                    self.array_elements.ptr[i].free()
```
Akuli:
```suggestion
                for elem = self.array_elements.ptr; elem < self.array_elements.end(); elem++:
                    elem.free()
```

compiler/constants.jou
```diff
@@ -30,8 +30,9 @@ enum ConstantKind:
     Float
     Double
     PointerString
-    ArrayOfBytes
     Bool
+    ArrayOfBytes  # TODO: delete this
```
Akuli:
```suggestion
    # ArrayOfBytes is a special case of Array for arrays of bytes. It is
    # needed for performance. For example, if a huge file is embedded with
    # embed_file(), we don't want to construct one Constant object for each
    # byte of the file.
    ArrayOfBytes
```

PR #1229, opened 2026-02-22
Create stdlib/json.jou

Akuli:
TODO:
- decide what to do with floats and locales (with `fi_FI.UTF-8` locale: `sprintf()`, `sscanf()` etc use `,` instead of `.` as decimal separator. Do we care if that breaks JSON?)
- document building JSON
- document parsing JSON
- test behavior of zero bytes inside JSON strings
- document behavior of zero bytes inside JSON strings
- document Infinity, -Infinity, NaN


Akuli:
I will split this up into mulitple smaller PRs.

PR #1231, opened 2026-02-22
Add JSON parsing to `stdlib/json.jou`

Akuli:
In #1230 I added a JSON builder to Jou stdlib. This PR adds a JSON parser.

Unlike other parsers, my JSON parser does not parse the whole JSON document into a data structure, but instead navigates it by passing around pointers to it.

Example (from docs in this PR):

```python
import "stdlib/json.jou"
import "stdlib/io.jou"
import "stdlib/mem.jou"

def main() -> int:
    json = "{\"number\": 123, \"nothing\": null, \"nope\": false, \"message\": \"hello\", \"array\": [1.2, 3.4, 5.6], \"object\": {\"a\":7, \"b\":8, \"c\":9}}"

    # Validating (may or may not be necessary depending on your use case)
    if json_is_valid(json):
        puts("is valid")  # Output: is valid

    # Getting a number
    printf("%f\n", json_to_double(json_get(json, "number")))  # Output: 123.000000

    # Getting null
    if json_is_null(json_get(json, "nothing")):
        puts("found the null")  # Output: found the null

    # Getting a boolean
    if json_is_true(json_get(json, "nope")):
        puts("this does not get printed")
    if json_is_false(json_get(json, "nope")):
        puts("this gets printed")  # Output: this gets printed

    # Getting a string
    message = json_to_string(json_get(json, "message"))
    puts(message)  # Output: hello
    free(message)

    # Looping through an array
    # Output: 1.200000
    # Output: 3.400000
    # Output: 5.600000
    for j = json_array_first(json_get(json, "array")); j != NULL; j = json_array_next(j):
        printf("%f\n", json_to_double(j))

    # Looping through an object
    # Output: a --> 7.000000
    # Output: b --> 8.000000
    # Output: c --> 9.000000
    for j = json_object_first(json_get(json, "object")); j != NULL; j = json_object_next(j):
        key = json_to_string(j)
        printf("%s --> %f\n", key, json_to_double(json_object_value(j)))
        free(key)

    return 0
```

### DIFF
--- a/tests/should_succeed/json_test_parsing.jou
+++ b/tests/should_succeed/json_test_parsing.jou
@@ -299,8 +299,8 @@ def test_numbers() -> None:
             # Windows likes to say e+029 instead of e+29
             s: byte[100]
             sprintf(s, "%g", json_to_double(valid_numbers[i]))
-            if strstr(s, "e+029") != NULL:
-                strcpy(strstr(s, "e+029"), "e+29")
+            if ends_with(s, "e+029"):
+                strcpy(strrchr(s, 'e'), "e+29")
             puts(s)
         else:
             printf("%g\n", json_to_double(valid_numbers[i]))


stdlib/json.jou
```diff
@@ -192,3 +196,505 @@ class JSONBuilder:
...
+#       \t
+#       \u followed by 4 hex digits (specifies 16-bit value as in UTF-16)
+#   '"'
+#
+# This function parses a string in JSON.
+#
+# Returns the JSON string s advanced just beyond the ending '"', or NULL if the
+# given JSON is invalid.
+def parse_string(s: byte*, ctx: StringParsingContext*) -> byte*:
+    if s == NULL or *s++ != '"':
```
Akuli: Perhaps this function shouldn't check for `NULL` either.

PR #1244, opened 2026-02-28
Add `JSONBuilder.string_with_len()` method

Akuli:
This way it is possible to build JSON that contains zero bytes. For example:

```python
import "stdlib/io.jou"
import "stdlib/json.jou"
import "stdlib/mem.jou"

def main() -> int:
    hello0world = ['h', 'e', 'l', 'l', 'o', '\0', 'w', 'o', 'r', 'l', 'd']
    jb = JSONBuilder{}
    jb.string_with_len(hello0world, 11)
    json = jb.finish()
    puts(json)  # Output: "hello\u0000world"
    free(json)
    return 0
```

### DIFF
--- a/tests/should_succeed/json_test_builder.jou
+++ b/tests/should_succeed/json_test_builder.jou
@@ -163,7 +163,7 @@ def test_strings() -> None:
     # Invalid UTF-8 is passed through as is. The check for U+2028/U+2029
     # shouldn't read beyond the end of the string even in this case.
     #
-    # Using malloc() so any problems can be found with valgrind.
+    # Using malloc() so any problems can hopefully be found with valgrind.
     s: byte* = malloc(2)
     s[0] = '\xe2'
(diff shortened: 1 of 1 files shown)


stdlib/json.jou
```diff
@@ -127,49 +128,55 @@ class JSONBuilder:
...
+    def string(self, s: byte*) -> None:
+        self.string_with_len(s, 0 if s == NULL else strlen(s))
 
     # This method is used for .string() and .key() methods.
-    def add_string_to_output(self, s: byte*) -> None:
+    def add_string_to_output(self, s: byte*, len: intnative) -> None:
         self.output.append('"')
-        while *s != '\0':
-            match *s:
+        i = 0
```
Akuli: Probably not necessary to support strings longer than 2GB, but just in case someone needs that..

```suggestion
        i: intnative = 0
```

PR #1260, opened 2026-03-03
Create stdlib/toml.jou, a TOML parser

Akuli:
I made a TOML 1.1.0 parser in Jou. This PR adds it to Jou standard library. It passes all [official TOML tests](https://github.com/toml-lang/toml-test/tree/v2.1.0) except anything that involves dates or times.

Example:

```python
import "stdlib/toml.jou"
import "stdlib/io.jou"

def main() -> int:
    toml = parse_toml("message = 'hello'")

    if toml.type == TOMLType.Error:
        toml.print()
        return 1

    message = toml_to_string(toml_get(&toml, "message"))
    if message != NULL:
        printf("it says %s\n", message)  # Output: it says hello

    toml.free()
    return 0
```

TODO:
- [x] rename methods so they make sense, e.g. `.simple_key()` --> `.dotted_key_component()`
- [x] add some way to debug print a `TOMLValue`
- [x] API change: the succeessfully parsed thing should be `TOMLValue` for convenience, not `List[TOMLKeyVal]` for minimalism
- [x] API change: rename `TOMLValueKind` to `TOMLType` and change the field accordingly
- [x] add recursion limit so it's not possible to overflow stack
- [x] ~~add function to read TOML from file (with size limit)~~
- [x] add a nicer way to find a specific value from the TOML, similar to `json_get()`
- [x] documentation!
- [x] set up GitHub Actions to run `tomltest.sh`
- [x] set up GitHub Actions to run `tomltest.sh` with valgrind
- [x] delete the temporary "valgrind everything" CI configuration
- [x] document the depth limit

### DIFF
--- a/stdlib/toml.jou
+++ b/stdlib/toml.jou
@@ -977,7 +977,7 @@ def toml_to_double(toml: TOML*) -> double:
         case _:
             return NAN
 
-# Unlike with json_to_string(), you should NOT need to free() the result!
+# Unlike with json_to_string(), you should NOT free() the result!
 @public
 def toml_to_string(toml: TOML*) -> byte*:
     if toml == NULL or toml.type != TOMLType.String or strlen(toml.string) != toml.string_len:


tests/toml/README.md
```diff
@@ -0,0 +1,62 @@
+# TOML Test Suite
+
+This directory contains scripts and configuration to test Jou's `stdlib/toml.jou`
```
Akuli:
```suggestion
This directory contains everything needed to test Jou's `stdlib/toml.jou`
```

tests/toml/run.sh
```diff
@@ -0,0 +1,196 @@
...
+# TOML test suite: https://github.com/toml-lang/toml-test
+
+set -e -o pipefail
+
+cd "$(dirname "$0")"
+
+RED="\x1b[31m"
+RESET="\x1b[0m"
+
+if [ "$(uname)" != "Linux" ] || [ "$(uname -m)" != "x86_64" ]; then
```
Akuli:
```suggestion
if [ "$(uname -sm)" != "Linux x86_64" ]; then
```

PR #1325, opened 2026-03-24
joutest: Support custom kinds of output comments instead of just `# Output: foo`

Akuli:
You can now define e.g. `# Error: foo` or `# MyCustomThing: foo` comments in `joutest.toml`. It is flexible enough to cover Jou compiler and joutest error messages.

### DIFF
--- a/tests/tests_of_joutest/wrong_type/expected_exit_code/joutest.toml
+++ b/tests/tests_of_joutest/wrong_type/expected_exit_code/joutest.toml
@@ -1,3 +1,0 @@
-[[tests]]
-files = "foo.jou"
-expected_exit_code = "1"  # Error: 'expected_exit_code' must be an integer between 0 and 255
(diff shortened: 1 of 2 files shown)


tools/joutest/step5_gather_expected_outputs.jou
```diff
@@ -2,41 +2,92 @@ import "stdlib/errno.jou"
...
+                params: byte*[enum_count(FormatParam)]
+                memset(params, 0, sizeof(params))
+                params[FormatParam.File as int] = test.path_in_working_directory
+                params[FormatParam.Line as int] = lineno_str
+                params[FormatParam.Comment as int] = comment
+
+                s = format_string(template, params)
+                out.extend_from_ptr(s, strlen(s))
+                out.append('\n')
+                free(s)
```
Akuli:
```suggestion
                free(s)
                return
```

PR #1329, opened 2026-03-26
Compiler refactoring: use just one builder per file

Akuli:
Before `AnyBuilder` was typically invoked 3 times per compilation: building LLVM IR, building the hash that goes to `.o` file name, and building UVGs. This PR rearranges the code so that all three are built at once. This is a step towards doing just one walk through the AST that includes type-checking and buliding (#1220).

### DIFF
--- a/compiler/main.jou
+++ b/compiler/main.jou
@@ -443,7 +443,7 @@ def main(argc: int, argv: byte**) -> int:
         case CompilingMode.ParseOnly:
             parse_only()
         case CompilingMode.UvgOnly:
-            check()  # This is enough, UVG builder knows when to print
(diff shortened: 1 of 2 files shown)


compiler/main.jou
```diff
@@ -451,7 +443,7 @@ def main(argc: int, argv: byte**) -> int:
         case CompilingMode.ParseOnly:
             parse_only()
         case CompilingMode.UvgOnly:
-            uvg_only()
+            check()  # This is enough, UVG builder knows when to print
```
Akuli:
```suggestion
            check()  # This is enough, UVG's will be printed as needed
```

PR #1353, opened 2026-04-10
joutest: Ignore SIGPIPE

Akuli:
At some point I managed to crash `joutest` with a child process that does not read any input, although I don't remember how I did it and cannot reproduce it now. In any case, this PR should fix that by ignoring SIGPIPE in joutest.

### DIFF
--- a/tools/joutest/step6_run_tests.jou
+++ b/tools/joutest/step6_run_tests.jou
@@ -789,6 +789,10 @@ if not WINDOWS:
     declare setenv(name: byte*, value: byte*, overwrite: int) -> int
     declare dirname(path: byte*) -> byte*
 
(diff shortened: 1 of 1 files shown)


tools/joutest/step6_run_tests.jou
```diff
@@ -803,23 +803,33 @@ if not WINDOWS:
...
 
 @public
 def step6_run_tests(tests: List[Test], args: CommandLineArgs*) -> None:
     if not WINDOWS:
-        old_path = getenv("PATH")
-        if old_path == NULL:
-            fail("no PATH environment variable")
-        old_path = strdup(old_path)
         add_joutest_dir_to_PATH()
+    #    ignore_sigpipe()
```
Akuli:
```suggestion
        ignore_sigpipe()
```

PR #1356, opened 2026-04-10
Allow casting between pointer and intnative

### LINKED ISSUE #1127: Casting 32-bit pointer to 32-bit integer
Currently Jou supports `some_pointer as int64`. On 32-bit platforms, IMO we should also support `some_pointer as int32`. But on 64-bit platforms that must be an error, because you would basically get a garbage value.

I think it makes sense for this to be a separate issue rather than a part of #304, because it is not absolutely essential for this to work for using Jou on a 32-bit platform.

Akuli:
Fixes #1127 

This only affects 32-bit systems.

You can now do e.g. `&foo as intnative` or `some_intnative_value as int*` on all supported platforms. Casting between a pointer and `int64` will soon be an error on 32-bit systems, but for now, both are allowed on 32-bit systems.


Akuli:
Seems like this can't be done in one step, because compiler/parser.jou casts pointers to integers. We need to first allow both, then update the bootstrap compiler to a version that supports both, then disallow int64 on 32-bit platforms.

PR #1368, opened 2026-04-30
bootstrap_transpiler.py: reuse operators as AST node names

Akuli:
This makes the bootstrap transpiler shorter and is a better fit for its "quick and dirty" style.

### DIFF
--- a/bootstrap_transpiler.py
+++ b/bootstrap_transpiler.py
@@ -2557,7 +2557,6 @@ class CFuncMaker:
                 lhs = self.do_expression(lhs_ast, None)
                 rhs = self.do_expression(rhs_ast, None)
                 result = self.add_variable(self.guess_type(expr))
-                op = "&" if op == "&" else "|" if op == "|" else "^"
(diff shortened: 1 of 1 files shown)


bootstrap_transpiler.py
```diff
@@ -2613,11 +2553,11 @@ def do_expression(self, expr, expected_type: JouType | None) -> JouValue:
                 self.output.append("}")
                 return result
 
-            case ("bit_and" | "bit_or" | "bit_xor" | "bit_shift_left" | "bit_shift_right" as op, lhs_ast, rhs_ast):
+            case ("&" | "|" | "^" | "<<" | ">>" as op, lhs_ast, rhs_ast):
                 lhs = self.do_expression(lhs_ast, None)
                 rhs = self.do_expression(rhs_ast, None)
                 result = self.add_variable(self.guess_type(expr))
-                op = "&" if op == "bit_and" else "|" if op == "bit_or" else "^"
+                op = "&" if op == "&" else "|" if op == "|" else "^"
```
Akuli:
```suggestion
```

Issue #1379, opened 2026-05-30
Explicit integer division operator


Akuli:
Currently Jou's division works like C (and Python 2):
- `8 / 3` produces `2` (floor division)
- `8 / 3.0` produces `2.666666...`
- There is no `//` operator

I think we should switch to Python 3 style division:
- `8 / 3` produces `2.666666...`
- `8 / 3.0` produces `2.666666...`
- `8 // 3` produces `2`

This way it is not possible to accidentally get the wrong result when trying to e.g. compute an average: `(4 + 7)/2` should be 5.5, not 5.

Akuli:
The `//` operator now exists, and using `/` between integers produces `compiler warning for file "a.jou", line 4: please use '//' for integer division, '/' will soon produce a double`

After a while, I will change the meaning of `/` so that `integer / integer` produces `double`, and delete the warning.

PR #1381, opened 2026-05-30
Support using `//` (double slash) for integer division

Akuli:
This PR adds a `//` operator that works just like `/`, but only accepts integers. This is the first part of #1379.

Example:

```python
import "stdlib/io.jou"

def main() -> int:
    printf("%d\n", 7 / 3)  # Output: 2
    printf("%d\n", 7 // 3)  # Output: 2
    printf("%f\n", 7.0 / 3)  # Output: 2.333333
    #printf("%f\n", 7.0 // 3)  # would be error
    return 0
```

If I uncomment the last line, I get:

```
compiler error in file "a.jou", line 7: wrong types: cannot floor divide double and int
```

Once this operator exists, the plan is to change existing Jou code to use it with the help of a compiler warning, and eventually change `integer / integer` so that it produces a double.

### DIFF
--- a/compiler/builder.jou
+++ b/compiler/builder.jou
@@ -1043,6 +1043,7 @@ class Builder:
     @public
     def unsigned_floor_div(self, a: BuilderValue, b: BuilderValue) -> BuilderValue:
         assert a.type == b.type
+        assert a.type.kind == TypeKind.UnsignedInteger
         result = BuilderValue{type = a.type, uvalue = ANONYMOUS_UVG_VALUE_ID}
         if self.llvm_builder != NULL:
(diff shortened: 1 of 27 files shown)


compiler/builder.jou
```diff
@@ -1039,24 +1039,34 @@ class Builder:
 
         return result
 
-    # a / b
+    # a // b, a and b are unsigned
     @public
-    def div(self, a: BuilderValue, b: BuilderValue) -> BuilderValue:
+    def unsigned_floor_div(self, a: BuilderValue, b: BuilderValue) -> BuilderValue:
         assert a.type == b.type
```
Akuli, line 1045: I can put more asserts here.

PR #1394, opened 2026-06-16
Add `is_ascii_string()` function to stdlib/ascii.jou

Akuli:
Reverts Akuli/jou#1268 and changes function name

I now have a use case for the `is_ascii_string()` function. In Windows API, there are two versions of many functions:
- `SomeFunctionA()` works with ASCII strings (and some other characters, but not anything nice like UTF-8)
- `SomeFunctionW()` works with all characters but expects input in a funny format that needs converting.

For ASCII strings, it's possible to avoid the conversion, which could be slightly faster, or at least simpler to debug and understand.

### DIFF
--- a/stdlib/ascii.jou
+++ b/stdlib/ascii.jou
@@ -2,6 +2,14 @@ import "stdlib/str.jou"
 import "stdlib/mem.jou"
 import "stdlib/list.jou"
 
+# Test if a string is ASCII only.
+@public
+def is_ascii(s: byte*) -> bool:
+    for p = s; *p != '\0'; p++:
+        if *p >= 128:
+            return False
+    return True
+
 # Check for '0', '1', ..., '9'.
 @public
 def is_ascii_digit(b: byte) -> bool:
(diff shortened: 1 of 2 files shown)


stdlib/ascii.jou
```diff
@@ -2,6 +2,14 @@ import "stdlib/str.jou"
 import "stdlib/mem.jou"
 import "stdlib/list.jou"
 
+# Test if a string is ASCII only.
+@public
+def is_ascii(s: byte*) -> bool:
```
Akuli: Nothing in the function name hints that this takes a string rather than a byte:

```suggestion
def is_ascii_string(s: byte*) -> bool:
```

tests/should_succeed/ascii_test.jou
```diff
@@ -6,6 +6,11 @@ import "stdlib/str.jou"
 import "stdlib/list.jou"
 
 def main() -> int:
+    assert is_ascii("hello")
+    assert is_ascii("hello world test!!! @$!")
+    assert is_ascii("")
+    assert not is_ascii("örkkimörkki")
```
Akuli:
```suggestion
    assert is_ascii_string("hello")
    assert is_ascii_string("hello world test!!! @$!")
    assert is_ascii_string("")
    assert not is_ascii_string("örkkimörkki")
```

